PR #126, opened 2020-03-23
fix: remove pro fontawesome, prettify, and fix linting

yousifalraheem:
- rename `seb-` prefix to `sebng-`
- remove pro fontawesome
- prettify
- fix linting


.prettierrc
```diff
@@ -1,7 +1,7 @@
-printWidth: 80
-singleQuote: true
+printWidth: 140
+singleQuote: false
```
omarVodiaK: Please don't :'(

yousifalraheem: What? you wanna keep single quotes? But it doesn't go with any other project we have currently.

mario-subo: YES! Death to single quotes!!! Muahahahaha 

just joking I don't care 👅

lib/src/core/selection/selection.ts
```diff
@@ -6,202 +6,198 @@
...
+    hasValue(): boolean {
+        return !this.isEmpty();
+    }
+
+    /**
+     * Sorts the selected values based on a predicate function.
+     */
+    sort(predicate?: (a: T, b: T) => number): void {
+        if (this._multiple && this.selected) {
+            this._selected && this._selected.sort(predicate);
```
omarVodiaK: i beliece we should start using this syntax
```suggestion
           this._selected?.sort(predicate);
```

yousifalraheem: I know but the Typescript version we have in this project doesn't recognize this syntax. I don't want to go through the process of updating Angular at the moment. At least not in this PR.

.prettierrc
yousifalraheem: This is very helpful when you're moving a lot of code around. All our old code is written with double-quotes. It means you're gonna deal with a lot of unnecessary linting issues.

projects/ng-components-docs/src/app/components/header/header.component.html
```diff
@@ -1,40 +1,43 @@
...
-        <div class="col-auto px-0">
-          <app-logo></app-logo>
+    <div class="row justify-content-between justify-content-lg-end h-100">
+        <div class="col col-md-auto d-none d-md-flex brand">
+            <div class="row mobile-header">
+                <div class="col-auto px-0">
+                    <app-logo></app-logo>
+                </div>
+                <a class="col-auto align-self-center text-center px-3 text-light" href="#" (click)="toggleMenu($event)">
+                    Menu
```
omarVodiaK: What will happen to the icon ?

lib/src/core/selection/selection.ts
omarVodiaK: I've seen this syntax in some other file lib/src/modal/modal.ts

PR #140, opened 2020-03-25
feat(radio-group): adds radio-group-component

### LINKED ISSUE #147: Add radio group component

mario-subo:
captured

fix #147

TODO: 
- [x] Add Inline and Description support
- [x] Add Unit Tests
- [x] Update style and add condensed
- [x] Add Disabled style
- [x] Add Form Control example
- [x] Fix style when element focused
- [x] Add keyboard support
- [x] Add support for custom HTML for labels


mario-subo:
Anyone already figured out how to make the API tab work? 
image

mario-subo:
> Anyone already figured out how to make the API tab work?
> image

fixed in re #141

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,137 @@
+import { Component, Input, forwardRef, OnChanges, SimpleChanges, Provider } from "@angular/core";
+import { NG_VALUE_ACCESSOR, ControlValueAccessor } from "@angular/forms";
+
+export interface Item {
```
yousifalraheem: Please rename this to `RadioGroupItem`

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    // HELPER ARRAYS
+    /** array of dropdown item elements with a unique id, the original optionItem and calculated selected property */
+    public uniqueList: Array<UniqueItem> = [];
+    /** Array of dropdown item elements which should be displayed in the current render cycle */
+    public displayList: Array<DisplayItem> = [];
+    /** Array of all dropdown item which are currently selected */
+    public selectedList: Array<Item> = [];
+
+    constructor() {}
```
yousifalraheem: You can remove this

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,137 @@
...
+            this._generateHelperArrays();
+        }
+    }
+
+    /** internal generate helper array function. Should be run on every change where the helper arrays need to be regenerated */
+    private _generateHelperArrays(): void {
+        this.uniqueList =
+            this.list &&
+            this.list
+                .filter(e => e && e.hasOwnProperty("key") && e.hasOwnProperty("value"))
```
yousifalraheem: add type for `e`

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,137 @@
...
+                .map((e: Item, i: number) => {
+                    const id = `${e.key}-${i}`;
+                    let selected = false;
```
yousifalraheem:
```suggestion
                    const id: string = `${e.key}-${i}`;
                    let selected: boolean = false;
```

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,137 @@
...
+        this.displayList =
+            this.uniqueList &&
+            this.uniqueList.map((e: UniqueItem) => {
+                return {
+                    ...e,
+                    className: `custom-form-radio${e.selected ? " selected" : ""}`,
+                };
+            });
+
+        this.selectedList = this.uniqueList && this.uniqueList.filter(e => e.selected).map(e => e.optionItem);
```
yousifalraheem: add type for `e`

projects/ng-components-docs/src/app/examples/components/radio-group/examples/radio-groups/radio-groups.component.ts
```diff
@@ -0,0 +1,16 @@
...
+})
+export class RadioGroupsComponent implements OnInit {
+    constructor() {}
+    list: Item[] = [
+        { key: "One", value: "1", label: "One" },
+        { key: "Two", value: "2", label: "Two" },
+    ];
+
+    ngOnInit() {}
+}
```
yousifalraheem:
```suggestion
import { Component } from "@angular/core";
import { Item } from "lib/src/radio-group/radio-group.component";

@Component({
    selector: "app-radio-groups",
    templateUrl: "./radio-groups.component.html",
})
export class RadioGroupsComponent implements OnInit {
    list: Item[] = [
        { key: "One", value: "1", label: "One" },
        { key: "Two", value: "2", label: "Two" },
    ];
}
```

mario-subo: you forgot to remove implements OnInit ;)

lib/src/radio-group/radio-group.component.html
```diff
@@ -1,15 +1,35 @@
...
+    <div class="{{ item.className }}" (click)="!(item.optionItem.disabled || disabled) ? handleItemOnClick(item) : null">
+        <input
+            [name]="id"
+            [id]="item.id"
+            [defaultChecked]="item.selected"
+            [disabled]="item.optionItem.disabled || disabled"
+            type="radio"
+            class="custom-control-input"
+        />
+        <label [attr.for]="!(item.optionItem.disabled || disabled) ? item.id : null" class="custom-control-label pr-1"
```
yousifalraheem: Please don't use `pr-1` since it introduces `!important`. Just add the padding to the stylesheet if necessary
```suggestion
        <label [attr.for]="!(item.optionItem.disabled || disabled) ? item.id : null" class="custom-control-label"
```

lib/src/radio-group/radio-group.component.html
```diff
@@ -1,15 +1,35 @@
...
+            >{{ item.optionItem.label }}
+            <span class="text-muted" *ngIf="item.optionItem.description">{{ item.optionItem.description }}</span>
```
yousifalraheem: Please don't use `text-muted` since it introduces `!important`. Just add the padding to the stylesheet if necessary
```suggestion
            <span *ngIf="item.optionItem.description">{{ item.optionItem.description }}</span>
```

lib/src/radio-group/radio-group.component.html
```diff
@@ -0,0 +1,35 @@
+<label *ngIf="label" class="radio-group-label"
+    ><b>{{ label }}</b></label
+>
+<div [attr.id]="id" *ngIf="list" class="radio-group {{ className }}" [class.disabled]="disabled">
```
yousifalraheem: In case there is no `className` passed as an input, this is how it's gonna look like `class="radio-group "` with a space at the end. I think if you can create a getter that appends a space with the `className` variable at the end would be a much cleaner solution. Don't you think?

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,139 @@
...
+    styleUrls: ["./radio-group.component.scss"],
+    providers: [CUSTOM_RADIOGROUP_CONTROL_VALUE_ACCESSOR],
+})
+export class RadioGroupComponent implements ControlValueAccessor, OnChanges {
+    @Input() list: Array<RadioGroupItem>; // TODO: Add support for custom html as well as string labels?
+    @Input() id?: string;
+    @Input() label?: string;
+    @Input() className?: string;
+    @Input() disabled?: boolean = false;
+    @Input() inline?: boolean = false;
```
yousifalraheem: The list of inputs here doesn't match the ones in [react-components](https://github.com/sebgroup/react-components/blob/develop/src/RadioGroup/RadioGroup.tsx#LC12). Please make sure we have the same amount of features. 

Please note, that we moved away from using `id` as the name of the input. Additionally, you need to add an implementation of random IDs. Please install `@sebgroup/frontend-tools` to access the `randomId` method.

lib/src/radio-group/radio-group.component.html
```diff
@@ -0,0 +1,35 @@
...
+        <ng-container *ngFor="let item of displayList; index as i">
+            <ng-container *ngTemplateOutlet="readioitem; context: { $implicit: item }"></ng-container>
+        </ng-container>
+    </ng-template>
+</div>
+
+<ng-template #readioitem let-item>
+    <div class="{{ item.className }}" (click)="!(item.optionItem.disabled || disabled) ? handleItemOnClick(item) : null">
+        <input
+            [name]="id"
```
yousifalraheem: replace this with name input

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    // HELPER ARRAYS
+    /** array of dropdown item elements with a unique id, the original optionItem and calculated selected property */
+    public uniqueList: Array<UniqueItem> = [];
+    /** Array of dropdown item elements which should be displayed in the current render cycle */
+    public displayList: Array<DisplayItem> = [];
+    /** Array of all dropdown item which are currently selected */
+    public selectedList: Array<RadioGroupItem> = [];
+
+    ngOnChanges(changes: SimpleChanges) {
```
DanSallau: ngOnChanges here is angular generic and will executed regardless of which property changed, why don't you use getter and setter functions for list instead ? I suggest moving this to setter function of list.

projects/ng-components-docs/src/app/examples/components/radio-group/examples/radio-groups/radio-groups.component.html
```diff
@@ -0,0 +1,25 @@
+<!--
+title: Radio Group
+componentid: component-radio-group
+variantid: component-radio-group-all
+url: https://sebgroup.github.io/ng-components/components/radio-group/examples
+-->
+
+<div class="row">
+    <div class="col">
+        <sebng-radio-group [id]="'simple'" [label]="'Simple'" [list]="list"></sebng-radio-group>
```
DanSallau: `Simple` here is a string right ?If so, Is there any reason why you are binding it to label instead of just passing it the usual `label="Simple"` ?

projects/ng-components-docs/src/app/examples/components/radio-group/examples/radio-groups/radio-groups.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component } from "@angular/core";
+import { RadioGroupItem } from "lib/src/radio-group/radio-group.component";
+import { FormControl } from "@angular/forms";
+
+@Component({
+    selector: "app-radio-groups",
+    templateUrl: "./radio-groups.component.html",
+})
+export class RadioGroupsComponent {
+    list: RadioGroupItem[] = [
```
DanSallau: Why not ?
```
    list: Array<RadioGroupItem> = [
```

lib/src/radio-group/radio-group.component.ts
yousifalraheem: I think he needs it to verify if there are changes, hence this condition
```typescript
if (changes.list && changes.list.previousValue !== changes.list.currentValue) {
```

projects/ng-components-docs/src/app/examples/components/radio-group/examples/radio-groups/radio-groups.component.ts
mario-subo: I'm not a fan of that. `RadioGroupItem[]` is easier to read and takes less space. Imagine you have `Array<Array<Item>>` ... this would be `Item[][]` instead

projects/ng-components-docs/src/app/examples/components/radio-group/examples/radio-groups/radio-groups.component.html
mario-subo: I've experienced issues with `prop="string"` on Angular 9. I try to do `[props]="'string'"` instead now

lib/src/radio-group/radio-group.component.ts
mario-subo: Yousif has a point however I think it's better to move it to a getter, you're right Nuru.

lib/src/radio-group/radio-group.component.html
mario-subo: I agree, I will use ngClass instead

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        return this._selectedValue;
+    }
+
+    // HELPER ARRAYS
+    /** array of dropdown item elements with a unique id, the original optionItem and calculated selected property */
+    public uniqueList: Array<UniqueItem> = [];
+    /** Array of dropdown item elements which should be displayed in the current render cycle */
+    public displayList: Array<DisplayItem> = [];
+    /** Array of all dropdown item which are currently selected */
+    public selectedList: Array<RadioGroupItem> = [];
```
yousifalraheem, line 98: why do the comments say dropdown?

mario-subo: woops lol

lib/src/radio-group/radio-group.component.ts
mario-subo: As discussed on slack we will not be using randomIds yet

lib/src/radio-group/radio-group.component.html
```diff
@@ -0,0 +1,48 @@
...
+                <ng-container *ngIf="condensed">
+                    <ng-container *ngTemplateOutlet="radioItem; context: { $implicit: item }"></ng-container>
+                </ng-container>
+            </div>
+        </ng-container>
+    </div>
+
+    <ng-template #noninline>
+        <ng-container *ngFor="let item of displayList; index as i">
+            <div *ngIf="!condensed" style="padding: 0.625rem;">
```
yousifalraheem, line 22: Move the styles from here to scss file

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+    private _classList: NgClass["ngClass"] = "radio-group";
+
+    @Input() disabled?: boolean = false;
+    @Input() condensed?: boolean = false;
+    @Input() inline?: boolean = false;
+
+    get classList(): NgClass["ngClass"] {
+        return this._classList;
+    }
```
yousifalraheem, line 68: I don't understand why won't you just use `ngClass` in the template, it's more than enough. 👇 
```html
<div *ngIf="list" class="radio-group" [ngClass]="className" [class.disabled]="disabled" [class.condensed]="condensed">
```

Even if you wanted to clean it up using Typescript, you can just do this instead:

```suggestion
    @Input() disabled?: boolean = false;
    @Input() condensed?: boolean = false;
    @Input() inline?: boolean = false;
    @Input() className?: string;

    get classList(): Array<string> {
        return ["radio-group", className];
    }
```

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    // Placeholders for the callbacks which are later provided
+    // by the Control Value Accessor
+    private onTouchedCallback: () => void;
+    private onChangeCallback: (_: any) => void;
+
+    private _selectedValue: RadioGroupItem = null;
+    public allSelected = null;
```
yousifalraheem, line 76: missing type

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,156 @@
...
+            this.onChangeCallback && this.onChangeCallback(state);
+            this.onTouchedCallback && this.onTouchedCallback();
+
+            this._generateHelperArrays();
+        }
+    }
+
+    get selectedValue() {
+        return this._selectedValue;
+    }
```
yousifalraheem, line 90: This getter doesn't do anything special, why don't you just use `selectedValue` as a variable?

lib/src/radio-group/radio-group.component.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    it("should create", () => {
+        expect(component).toBeTruthy();
+    });
+
+    it("Should render label", async(() => {
+        component.label = "Country";
+        fixture.detectChanges();
+        const label = fixture.debugElement.query(By.css(".radio-group-label > b"));
+        expect(label).not.toBeNull();
+        expect(label.nativeElement.innerHTML).toEqual("Country");
```
yousifalraheem, line 67: I don't think the query returns `null` if it's not found, it will return `undefined` instead
```suggestion
        const label: HTMLLabelElement = fixture.debugElement.query(By.css(".radio-group-label"));
        expect(label).toBeDefined();
        expect(label.nativeElement.innerHTML).toEqual("<b>Country</b>");
```

lib/src/radio-group/radio-group.component.html
```diff
@@ -0,0 +1,48 @@
+<label *ngIf="label" class="radio-group-label"
+    ><b>{{ label }}</b></label
+>
+
+<div *ngIf="list" [ngClass]="classList" [class.disabled]="disabled" [class.condensed]="condensed">
+    <div *ngIf="inline; else noninline" class="row">
+        <ng-container *ngFor="let item of displayList; index as i">
+            <div class="col">
+                <div *ngIf="!condensed" style="padding: 0.625rem;">
```
yousifalraheem, line 9: Move the styles from here to scss file

lib/src/radio-group/radio-group.component.ts
mario-subo: actually allSelected should not be there as radio group is not multiple choice, I thought I removed it

lib/src/radio-group/radio-group.component.ts
mario-subo: you're right there used to be side effects in there but I cleaned that up, I will use a public var instead

mario-subo: actually after checking I can see there are side effects happening in the setter of selectedValue, so this is correct. the getter should just return the value. If I used selectedValue as a normal value I would have to implement ngOnChanges instead.

lib/src/radio-group/radio-group.component.spec.ts
mario-subo: you mean `const label: DebugElement`

mario-subo: Error: Expected '<b _ngcontent-a-c40=""'>Country</b>` to equal `<b>Country</b>`

mario-subo: @kherP any suggestions?

lib/src/radio-group/radio-group.component.ts
mario-subo: You're right this code is overkill. Was doing it to put all the classes in one list to keep it organized but it turned out not the way I hoped :)

lib/src/radio-group/radio-group.component.spec.ts
kherP: supposedly if not found, it will return null

kherP: and if it is null, `expect(null).toBeDefined()` will pass the test, i believe in this `not.toBeNull()` is better

yousifalraheem: @mario-subo 
Just use a regex `innerHTML.replace(/\<b ?(.*?)\>/g, "<b>")` I already updated the suggestion

mario-subo: but why not just use `"label > b"` and check the string itself. I don't understand the unnecessary complexity. What does omitting `> b` and adding regex actually solve?

yousifalraheem: Now I think about it, why do you have a `b` tag? if you're following bootstrap styles, a `label` tag doesn't need a `b` tag inside. If they chose to make it bold they will make it bold

See this example, we don't render `b` tag inside anymore
https://sebgroup.github.io/react-components/#/radiogroup

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,174 @@
+import { Component, Input, forwardRef, Provider, ViewChildren, QueryList, ElementRef, AfterViewChecked } from "@angular/core";
+import { NG_VALUE_ACCESSOR, ControlValueAccessor } from "@angular/forms";
+import { NgClass } from "@angular/common";
```
yousifalraheem, line 3: I think this is unused import now

lib/src/radio-group/radio-group.component.spec.ts
kherP: we can omit the ` > b ` by setting the expect to `expect(label.nativeElement.innerHTML).toContain("Country");`

I don't understand why we need the regex as well haha

mario-subo: yeah I will use your suggestion KP. Yousif as far as b inside the label goes I have been following that because the design used to be like that. I am thinking of removing the label altogether from all the components. I don't think it's needed. but as it stands now all my labels have a b in them so for now I will leave it and we can refactor all labels later.

lib/src/radio-group/radio-group.component.ts
mario-subo: fixed

lib/src/radio-group/radio-group.component.spec.ts
omarVodiaK: @KP suggestion is good, FYI you are already checking if b tag is a direct child of .radio-group-label here
const label = fixture.debugElement.query(By.css(".radio-group-label > b"));

lib/src/radio-group/radio-group.component.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export interface RadioGroupItem {
+    /** The label or text to be displayed in the list */
+    label?: string;
+    /**
+     * A Custom template as html string to be used instead of label and description.
+     *
+     * Example: "<b>Hello</b>"
+     * */
+    customTemplate?: string;
```
omarVodiaK, line 12: This can be a bit hard to understand, but i couldn't find a better way to do it, the only way i know is to declare this one as `TemplateRef<HTMLElement>` and then value you pass to it could be` "<span>test<span>" as unknown as HTMLElement`

projects/ng-components-docs/src/app/examples/components/radio-group/examples/radio-groups/radio-groups.component.ts
```diff
@@ -0,0 +1,33 @@
...
+    list4: RadioGroupItem[] = [...this.list];
+    list5: RadioGroupItem[] = [...this.list, { key: "Three", value: "3", label: "Three disabled", disabled: true }];
+    list6: RadioGroupItem[] = [...this.list];
+    list7: RadioGroupItem[] = [...this.list];
+    list8: RadioGroupItem[] = [
+        ...this.list,
+        {
+            key: "Three",
+            customTemplate:
+                "<div class='btn btn-outline-primary m-0'><h5>I'm a custom html element :)</h5> </ br> <p> You can put whatever you like in here </p></div>",
```
omarVodiaK, line 26: beautiful

PR #145, opened 2020-03-26
feat(textlabel component): textlabel component

omarVodiaK:


lib/src/textLabel/textLabel.component.html
```diff
@@ -0,0 +1,4 @@
+<div class="text-label" [ngClass]="className">
```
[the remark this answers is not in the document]
yousifalraheem: You can use bootstrap CSS variable, please don't add this class since it introduces `!important`

projects/ng-components-docs/src/app/examples/components/textLabels/text-labels.module.ts
```diff
@@ -0,0 +1,60 @@
...
+                            description: "Additional description for example (optional)",
+                            sources: [
+                                {
+                                    name: "text-labels.component.html",
+                                    src: require("!raw-loader!./text-labels.component.html"),
+                                    lang: "markup",
+                                },
+                                {
+                                    name: "text-labels.component.ts",
+                                    // @ts-ignore
```
yousifalraheem: why is this here? what's to be ignored?

projects/ng-components-docs/src/app/examples/components/textLabels/text-labels.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="route-template">
+    <div class="info-holder">
+        <div class="info">
+            <div class="md-file"></div>
+        </div>
+
+        <div class="info">
+            <h2>Output</h2>
```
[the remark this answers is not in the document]
yousifalraheem: Agree, we don't need the old classes since this uses Storybook, just check the other examples as a reference

lib/src/textLabel/textLabel.component.html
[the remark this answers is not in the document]
omarVodiaK: yea no way mb-1 will not be useful in this case.

lib/src/textLabel/textLabel.component.ts
```diff
@@ -0,0 +1,25 @@
...
+    encapsulation: ViewEncapsulation.None,
+})
+export class TextLabelComponent {
+    @Input() id?: string;
+    @Input() label?: string | TemplateRef<HTMLElement>;
+    @Input() value?: string | TemplateRef<HTMLElement>;
+    @Input() className?: string | Array<string> | { [key: string]: boolean };
+
+    constructor() {
+        console.log("here");
```
DanSallau: The constructor here will be removed right ?

lib/src/textLabel/textLabel.component.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    /**
+     * Check if input parameter is a string
+     */
+    isString = (input: any): boolean => {
+        return typeof input == "string";
```
DanSallau:
```suggestion
        return typeof input === "string";
```

lib/src/textLabel/textLabel.component.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    beforeEach(() => {
+        fixture = TestBed.createComponent(TextLabelComponent);
+        component = fixture.componentInstance;
+        fixture.detectChanges();
+    });
+
+    it("should create", () => {
+        expect(component).toBeTruthy();
+    });
+});
```
DanSallau: I think your test here didn't cover a scenario where either label or value is string.

lib/src/textLabel/textLabel.component.ts
mario-subo: Nice catch!

lib/src/textLabel/textLabel.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="text-label" [ngClass]="className" id="id">
+    <ng-container *ngTemplateOutlet="isString(label) ? defaultLabel : label"></ng-container>
+    <ng-container *ngTemplateOutlet="isString(value) ? defaultValue : value"></ng-container>
```
yousifalraheem, line 3: Don't you think we should replace string rendering with `ng-content`?

omarVodiaK: `*ngTemplateOutlet` gives us some more power of showing the default template if none is provided. You cannot conditionally render the content within `ng-content`. You have to show the content that is received from the parent with no means to make decisions based on the content, which is what we want to achieve now.

lib/src/textLabel/textLabel.component.spec.ts
omarVodiaK: added more unit tests

lib/src/textLabel/textLabel.component.ts
omarVodiaK: fixed

lib/src/textLabel/textLabel.component.ts
omarVodiaK: removed the constructor

lib/src/textLabel/textLabel.component.html
```diff
@@ -0,0 +1,4 @@
+<div class="text-label" [ngClass]="className">
+    <label *ngIf="label" class="custom-label" for="name">{{ label }}</label>
```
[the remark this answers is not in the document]
omarVodiaK: removed

lib/src/textLabel/textLabel.component.html
DanSallau: I understand that we can achieve conditional ng-content with *ngTemplateOutlet and from your comment above i can see another different between the two. But I am a little bit skeptical about adoption/implementation/usage though as far as users are concern . Do you thing the users are familiar with *ngTemplateOutlet  and it usage ?

PR #149, opened 2020-03-26
build(unit-tests): fixed config for tests

mario-subo:


package.json
```diff
@@ -79,6 +79,8 @@
         "@datorama/akita-ngdevtools": "^1.0.2",
         "@types/jasmine": "~2.8.8",
         "@types/jasminewd2": "~2.0.3",
+        "@types/jest": "^25.1.4",
+        "@types/mocha": "^7.0.2",
```
yousifalraheem: why do you need jest and mocha?

tsconfig.json
```diff
@@ -13,7 +13,7 @@
         "experimentalDecorators": true,
         "importHelpers": true,
         "target": "es2015",
-        "types": ["node"],
+        "types": ["jasmine", "jest", "mocha", "node"],
```
yousifalraheem: I think adding jasmine alone is enough, no need for jest and mocha. It's going to be confusing for maintainers

mario-subo: When I was testing I get errors if I don't add jest and mocha here. Can you provide a fix if you know how to do it?

package.json
mario-subo: because we use jest and mocha in tests

DanSallau: No, i don't think we ever used mocha and jest for angular. karma, jasmine etc. if my memory serve me right.

tsconfig.json
DanSallau: Can we transfer the config we have in our old Angular components which works btw to here ?

package.json
mario-subo: yeah this is a hack we have to find the underlying issue with tsconfig not finding types from jasmine

tsconfig.json
mario-subo: you can try but it's more or less the same here. Might be differences in how the config has to be setup because this is Angular 8 and the old one was Angular 7

PR #166, opened 2020-04-13
Feature/angular 9

mario-subo:
Upgrades the entire project to angular 9

passed scripts:
- [x] npm run docs
- [x] npm run test 
- [x] npm run build-lib
- [x] npm run build-docs


projects/ng-components-docs/tsconfig.app.json
```diff
@@ -11,5 +11,6 @@
             "path": ["lib/src/button/*"]
         }
     },
-    "exclude": ["test.ts", "**/*.spec.ts"]
+    "files": ["src/main.ts", "src/polyfills.ts"],
```
omarVodiaK, line 14: is it ok to remove exclude ?

mario-subo: this was done automatically by ng update script. looks like they are moving away from exclude to include or files instead. But to be honest I'm not sure. I need to test out all the build scripts etc

PR #172, opened 2020-04-14
Feature pagination

DanSallau:
pagination


projects/ng-components-docs/src/app/examples/components/pagination/examples/pagination/pagination.component.html
```diff
@@ -0,0 +1,8 @@
+<sebng-pagination
+    [value]="pagination"
+    (onChange)="setPagination($event)"
```
yousifalraheem, line 3: Angular doesn't like to use `on` prefix on methods and in a lot of cases, it will throw an error. Can you replace it with just `change`

lib/src/pagination/pagination.component.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.dotnavList = genList;
+        this.list = genList.slice(start, end);
+    }
+
+    ngOnChanges(changes: SimpleChanges) {
+        if (changes.offset || changes.size) {
+            this.setPagingSize();
+        }
+
+        if (changes.pagingLength || changes.value) {
```
yousifalraheem, line 73:
```suggestion
        if (changes.pagingLength !== this.pagingLength || changes.value !== this.value) {
```

lib/src/pagination/pagination.component.ts
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            end = start + length;
+        }
+        this.dotnavList = genList;
+        this.list = genList.slice(start, end);
+    }
+
+    ngOnChanges(changes: SimpleChanges) {
+        if (changes.offset || changes.size) {
```
yousifalraheem, line 69:
```suggestion
        if (changes.offset !== this.offset || changes.size !== this.size) {
```

lib/src/pagination/pagination.component.ts
```diff
@@ -0,0 +1,81 @@
...
+        const initialOffset: number = this.offset ? this.offset : 10;
+        this.pagingSize = Math.ceil(Number(this.size) / initialOffset);
+        this.generateList();
+    }
+    /**
+     * Generates an array of the pages that needs to be displayed
+     * It depends on the size, offset, and the current value
+     * @param {number} value The current value. The current page.
+     * @param {number} length The length of the pagination, number of pages to be displayed.
+     * @returns void
```
yousifalraheem, line 41: There are no params nor a returned value
```suggestion
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,99 @@
...
+                    </button>
+                </li>
+            </ng-container>
+        </ul>
+        <ng-container *ngIf="useDotNav">
+            <ul class="pagination dotnav">
+                <ng-container *ngFor="let num; dotnavList">
+                    <li class="page-item" [ngClass]="{ active: value === num }" (click)="handleOnChange(num)" value="num">
+                        <span class="page-dot-link" [ngClass]="{ active: value === num }"></span>
+                    </li>
```
yousifalraheem, line 94: Why are you setting class `active` in parent and child? The parent is enough. 

Don't forget the binding
```suggestion
                    <li class="page-item" [class.active]="value === num" (click)="handleOnChange(num)" [value]="num">
                        <span class="page-dot-link"></span>
                    </li>
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,99 @@
...
+                    </button>
+                </li>
+            </ng-container>
+            <ng-container *ngFor="let num of list; let i = $index">
+                <li class="page-item" [ngClass]="{ active: value === num }" (click)="handleOnChange(num)" value="num">
+                    <button class="page-link" [ngClass]="{ active: value === num }">
+                        <span class="nav-num">{{ num }}</span>
+                        <span class="sr-only">{{ num }}</span>
+                    </button>
+                </li>
```
yousifalraheem, line 49: No need for class `active` to be set for both parent and child. Remember you need to fix it in the CSS

Don't forget the binding as well
```suggestion
                <li class="page-item" [class.active]="value === num" (click)="handleOnChange(num)" [value]="num">
                    <button class="page-link">
                        <span class="nav-num">{{ num }}</span>
                        <span class="sr-only">{{ num }}</span>
                    </button>
                </li>
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,99 @@
+<div class="pagination-wrapper" [ngClass]="className" id="id">
```
yousifalraheem, line 1: Shouldn't this be one-way binding?
```suggestion
<div class="pagination-wrapper" [ngClass]="className" [id]="id">
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,99 @@
...
+                <li class="page-item" [ngClass]="{ active: value === num }" (click)="handleOnChange(num)" value="num">
+                    <button class="page-link" [ngClass]="{ active: value === num }">
+                        <span class="nav-num">{{ num }}</span>
+                        <span class="sr-only">{{ num }}</span>
+                    </button>
+                </li>
+            </ng-container>
+            <ng-container *ngIf="value !== pagingSize">
+                <li class="page-item" (click)="handleOnChange(value + 1)">
+                    <button class="page-link" title="nextText">
```
yousifalraheem, line 53: Don't forget the binding
```suggestion
                    <button class="page-link" [title]="nextText">
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,99 @@
...
+                                </svg>
+                            </ng-template>
+                        </span>
+                        <span class="sr-only">{{ firstText || "First" }}</span>
+                    </button>
+                </li>
+            </ng-container>
+            <ng-container *ngIf="value !== 1">
+                <li class="page-item" (click)="handleOnChange(value - 1)">
+                    <button class="page-link" title="previousText">
```
yousifalraheem, line 26: Don't forget the binding
```suggestion
                    <button class="page-link" [title]="previousText">
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,99 @@
+<div class="pagination-wrapper" [ngClass]="className" id="id">
+    <nav class="custom-pagination">
+        <ul class="pagination">
+            <ng-container *ngIf="value !== 1 && useFirstAndLast">
+                <li class="page-item" (click)="handleOnChange(1)">
+                    <button class="page-link" title="firstText">
```
yousifalraheem, line 6: Don't forget the binding
```suggestion
                    <button class="page-link" [title]="firstText">
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,99 @@
...
+                            </ng-template>
+                        </span>
+
+                        <span class="sr-only">{{ nextText || "Next" }}</span>
+                    </button>
+                </li>
+            </ng-container>
+            <ng-container *ngIf="value !== pagingSize && useFirstAndLast">
+                <li class="page-item" (click)="handleOnChange(pagingSize)">
+                    <button class="page-link" title="lastText">
```
yousifalraheem, line 71: Don't forget the binding
```suggestion
                    <button class="page-link" [title]="lastText">
```

projects/ng-components-docs/src/app/examples/components/pagination/pagination.module.ts
```diff
@@ -0,0 +1,57 @@
...
+                            description: "Additional description for example (optional)",
+                            sources: [
+                                {
+                                    name: "pagintaion.component.html",
+                                    src: require("!raw-loader!./examples/pagination/pagination.component.html"),
+                                    lang: "markup",
+                                },
+                                {
+                                    name: "pagination.component.ts",
+                                    src: require("!raw-loader!./examples/pagination/pagination.component.ts"),
```
yousifalraheem, line 34: I don't think you need to mention `raw-loader`

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,101 @@
+<div class="pagination-wrapper" [ngClass]="className" [id]="id || ''">
```
omarVodiaK, line 1: if id doesnt exist then dont give it an id, think about what will happen if another element have id=""
i suggest
```typescript
<div class="pagination-wrapper" [ngClass]="className" [attr.id]="id">
 ```
this syntax the id property will not be rendered if there is no id

File: lib/src/pagination/pagination.component.html [hunk 14]
```diff
@@ -0,0 +1,101 @@
...
+                        <span class="page-dot-link" [ngClass]="{ active: value === num }"></span>
+                    </li>
+                </ng-container>
+            </ul>
+        </ng-container>
+        <ng-template #numNav>
+            <ul class="pagination">
+                <ng-container *ngIf="value !== 1 && useFirstAndLast">
+                    <li class="page-item" (click)="handleOnChange(1)">
+                        <button class="page-link" [title]="firstText || ''">
```
omarVodiaK, line 16:
```typescript
[attr.title]="firstText"
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,101 @@
...
+                                    </svg>
+                                </ng-template>
+                            </span>
+                            <span class="sr-only">{{ firstText || "First" }}</span>
+                        </button>
+                    </li>
+                </ng-container>
+                <ng-container *ngIf="value !== 1">
+                    <li class="page-item" (click)="handleOnChange(value - 1)">
+                        <button class="page-link" [title]="previousText || ''">
```
omarVodiaK, line 36:
```typescript
[attr.title]="previousText"
```

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,101 @@
...
+                    <li class="page-item" [ngClass]="{ active: value === num }" (click)="handleOnChange(num)" value="num">
+                        <button class="page-link" [ngClass]="{ active: value === num }">
+                            <span class="nav-num">{{ num }}</span>
+                            <span class="sr-only">{{ num }}</span>
+                        </button>
+                    </li>
+                </ng-container>
+                <ng-container *ngIf="value !== pagingSize">
+                    <li class="page-item" (click)="handleOnChange(value + 1)">
+                        <button class="page-link" [title]="nextText || ''">
```
omarVodiaK, line 63:
```typescript
[attr.title]="nextText"
```

lib/src/pagination/pagination.component.ts
omarVodiaK:
```suggestion
        if ((changes.offset && changes.offset !== this.offset) || (changes.size && changes.size !== this.size)) {
```

lib/src/pagination/pagination.component.ts
omarVodiaK:
```suggestion
        if ((changes.pagingLength && changes.pagingLength !== this.pagingLength) || (changes.value && changes.value !== this.value)) {
```

lib/src/pagination/pagination.component.ts
omarVodiaK: in case the offset become undefined or null this changes.offset !== this.offset will be true

lib/src/pagination/pagination.component.html
yousifalraheem: I think if he initializes `id` with an empty string then it will be rendered, but if he leaves it as `undefined` or initializes it with `null` then it's not gonna be rendered

lib/src/pagination/pagination.component.ts
yousifalraheem: @omarVodiaK `if (changes.offset) {` will return false if the offset is `0` which is a valid value. Maybe it's best to use [isEmpty](https://github.com/sebgroup/frontend-tools/wiki/isEmpty) utility instead

lib/src/pagination/pagination.component.ts
```diff
@@ -0,0 +1,78 @@
...
+
+@Component({
+    selector: "sebng-pagination",
+    templateUrl: "./pagination.component.html",
+    styleUrls: ["./pagination.component.scss"],
+    encapsulation: ViewEncapsulation.None,
+})
+export class PaginationComponent implements OnChanges {
+    @Input() size: number;
+    @Output() handleChange: EventEmitter<any> = new EventEmitter();
```
yousifalraheem, line 11: you don't need this, you already have `handleOnChange`, just implement [ControlValueAccessor](https://angular.io/api/forms/ControlValueAccessor) which will enable `value` input and `change` emitter

lib/src/pagination/pagination.component.ts
DanSallau: You don't have to compare all these, ngOnChage lifecycle is smart enough to know that.

lib/src/pagination/pagination.component.ts
DanSallau: `changes.offset` is an object with props `currentValue` `firstChange` `previousValue`. I don't see the need for either of these checking in this case. Also you don't have to compare with the previous value. ngOnChage does that .

projects/ng-components-docs/src/app/examples/components/pagination/pagination.module.ts
DanSallau: I am not sure if it will be able to parse the file .ts to generate the api as intended. Neither am i aware of the consequences of removing the ra-loader. I don't know how this configuration handles the api generation, but going by the other examples, i will not be in a heist to parse it without the raw-loader.

lib/src/pagination/pagination.component.html
DanSallau: It looks like a cleaner solution.

lib/src/pagination/pagination.component.ts
DanSallau: I don't get it, its simple onChange, i simply rename to  handleChange. If any thing, i think the type of EventEmitter is wrong. Also why will you need ControlValueAccessor on a non input or controlled component ? I deviate from using change because it seems to be a reserve angular bind word and might be confusing.

lib/src/pagination/pagination.component.html
```diff
@@ -0,0 +1,101 @@
+<div class="pagination-wrapper" [ngClass]="className" [attr.id]="id">
+    <nav class="custom-pagination">
```
mario-subo, line 2: since there is no custom css for the pagination wrapper and custom pagination I see them as unnecessary wrappers which should be removed

lib/src/pagination/pagination.component.ts
yousifalraheem: `ngOnChanges` gets called after any little change, it doesn't know whether it's `pagingLength` or something else. You don't need to recalculate `setPagingSize()` when changes happen to something other than `pagingLength` or `value`. `changes` object is just the old values, something like `previousProps` and `this` is the `newProps`. You still need to compare them.

yousifalraheem: You're right, it's mentioned in their [docs](https://angular.io/guide/lifecycle-hooks#onchanges)

lib/src/pagination/pagination.component.ts
yousifalraheem: HTML doesn' have a full arsenal of customizable components, that's why we create **custom** ones. It's up to us to decide how it should behave. In this case, pagination has a value, and when it changes we need to reflect the change to the parent. It kind of behaves like a slider with a range of values. A slider is a form element. Therefore, we decided to make pagination behaves like a form element. With a `value` and `onChange` handler.

It is important to standardize implementations like this so that in the future we have standardized change handlers based on the `name` attribute.

yousifalraheem: I'm expecting the component to be used like this:
```html
<sebng-pagination
    offset="5"
    size="25"
    [value]="value"
    (change)="handleChanges($event)
></sebng-pagination>
```

lib/src/pagination/pagination.component.html
yousifalraheem: I think you should leave this class for now until I refactor the React version to use Bootstrap markup. Once I'm done, I'll be making the necessary changes. No need to waste time on it now

lib/src/pagination/pagination.component.ts
DanSallau: We can still that. I rename it .

PR #185, opened 2020-04-16
refactor(project): removes unused src folder

mario-subo:


src/tsconfig.spec.json
```diff
@@ -1,9 +0,0 @@
-{
```
DanSallau, line 1: You moved the config here to the main tsconfig ? I am concern about the jasmine and node types definition below. Not having them means the jasmine terms will not be recognized.

PR #207, opened 2020-04-22
feat(textboxcomponent): textboxGroup and tests

DanSallau:
textgroup


lib/src/textboxGroup/textboxGroup.component.html
```diff
@@ -0,0 +1,70 @@
+<div class="form-group input-box-group" [ngClass]="className">
+    <ng-container *ngIf="label">
+        <label class="custom-label" [htmlFor]="internalId">
+            {{ label }}
+        </label>
+    </ng-container>
+    <div class="input-group" [ngClass]="{ success: success, 'has-error': error }" [class.disabled]="disabled">
```
omarVodiaK, line 7:
```suggestion
    <div class="input-group" [ngClass]="{ success: success, 'has-error': error, disabled: disabled }">
```

DanSallau: I recognised and agree with above @omarVodiaK . But i refused to combine them because i feel like while there is a relationship between success and error, i feel like disabled is separated and wanted to preserve that distinction. This however is just a preference only. I will revert it back if it has any bearing on the feature or code quality.

lib/src/textboxGroup/textboxGroup.component.ts
```diff
@@ -0,0 +1,153 @@
...
+    @Output() onKeyUp = new EventEmitter<KeyboardEvent>();
+    @Output() onLeftClick = new EventEmitter<MouseEvent>();
+    @Output() onRightClick = new EventEmitter<MouseEvent>();
+
+    private innerValue: Value;
+
+    private onTouchedCallback: () => void;
+    private onChangeCallback: (_: any) => void;
+
+    public internalId: string;
```
omarVodiaK, line 72: i dont think this should be public

lib/src/textboxGroup/textboxGroup.component.ts
```diff
@@ -0,0 +1,153 @@
...
+    public displayError: boolean;
```
omarVodiaK, line 73: same here

lib/src/textboxGroup/textboxGroup.component.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    ngOnInit() {
+        this.internalId = this.id ? this.id : this.label ? randomId("tbg-") : null;
```
omarVodiaK, line 76: you are forcing the user to have an id, i dont think you should do that and instead use @Input() id directly in the component template

lib/src/textboxGroup/textboxGroup.component.html
omarVodiaK: they are all classNames.

lib/src/textboxGroup/textboxGroup.component.ts
DanSallau: I wouldn't say forcing the user, we are putting one by default for him incase he chose not to. I couldn't remember what the discussion was , but we had a discussion last time and decided to go with ramdom id for specific input elements of which texboxGroup was among.  I had remove it here, but i couldn't remember  what we address with it in react-component. We might have to discuss again perhaps.

lib/src/textboxGroup/textboxGroup.component.ts
DanSallau: They are not privates , we use them within the .html file.

yousifalraheem: These are not inputs and should not be exposed to the users, `private` would be a better declaration

lib/src/textboxGroup/textboxGroup.component.ts
yousifalraheem: I believe the random ID serves one purpose here, enabling a feature where clicking on the label will focus on the input. I don't think there is an issue with that.

lib/src/textboxGroup/textboxGroup.component.ts
DanSallau: No they are not private, if you make them private , then you can't access them outside the file especially inside the .html file.

lib/src/textboxGroup/textboxGroup.component.ts
mario-subo: He's using it in the HTML though so it has to be. Maybe it should be renamed.

lib/src/textboxGroup/textboxGroup.component.html
DanSallau: Ok the disabled kid is placed in the same class.

PR #247, opened 2020-05-11
feat(toggle-component): add new toggle component

DanSallau:


projects/ng-components-docs/src/app/examples/components/toggle/examples/toggle/toggle.component.html
```diff
@@ -0,0 +1,5 @@
+<div class="row">
+    <div class="col">
+        <sebng-toggle name="myToggle" label="Toggle label" [(ngModel)]="toggleValue" id="my-toggle" (onChange)="(onChange)"></sebng-toggle>
```
yousifalraheem, line 3: Why do you need both `ngModel` and `onChange` ??? shouldn't `ngModel` be enough?

DanSallau: onChange just like we discuss before does nothing special. Its just. callback to mimic the native onchange. So if i want to run a logic after a change, here is the place to do that. Alternatively i can use `ngModelChange` instead , but this requires me to use `[(ngModel)]` . So, i can just use `onChange` even without `ngModel` like the native change event here instead , execute my logic on every change.

yousifalraheem: I'm not saying you should remove it, I'm just saying why do you need it in the example, the `ngModel` should handle the two-way data binding

DanSallau: Ok that we can do without. I removed.

PR #248, opened 2020-05-11
feat(tabs): tabs component

DanSallau:


projects/ng-components-docs/src/app/examples/components/tabs/examples/tabs/tabs.component.html
```diff
@@ -0,0 +1,5 @@
+<div class="row">
+    <div class="col">
+        <sebng-tabs [list]="tabList" [activeTab]="activeTab" (onClick)="onClick($event)"></sebng-tabs>
```
mario-subo, line 3: I thought we said we are not naming our actions onClick because Angular is complaining about it?

DanSallau: Didn't know that's an issue, we had onChange in previous components and didn't se any issue. What was the angular complain ?

yousifalraheem: Angular usually complains if you use `onAnything` methods because it's usually reserved by Javascript. I don't think this is the last version of this component before releasing so that's Ok.

PR #253, opened 2020-05-12
feat(chip): new chip component

DanSallau:


projects/ng-components-docs/src/app/examples/components/chip/chip.component.html
```diff
@@ -0,0 +1,7 @@
+<div class="row">
+    <div class="col">
+        <sebng-chip [className]="'m-2'" (onClose)="onClose($event)">
```
yousifalraheem, line 3: I think you can do this instead
```suggestion
        <sebng-chip className="m-2" (onClose)="onClose($event)">
```

lib/src/chip/chip.component.html
```diff
@@ -0,0 +1,6 @@
+<div class="custom-chip" [ngClass]="className" [attr.id]="id">
+    <div class="content"><ng-content></ng-content></div>
```
mario-subo, line 2: Please use `select` to project the content. It should look like this. 
`<ng-content select="[chipContent]"></ng-content>`

https://stackoverflow.com/questions/52638718/multiple-ng-content
image

lib/src/chip/chip.component.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { async, ComponentFixture, TestBed } from "@angular/core/testing";
+
+import { ChipComponent } from "./chip.component";
+import { By } from "@angular/platform-browser";
+import { Component } from "@angular/core";
+
+@Component({
+    selector: "test-sebng-chip",
+    template: ` <sebng-chip [id]="id" [className]="className" (onClose)="onClose($event)">Test</sebng-chip> `,
```
mario-subo, line 9:
```suggestion
    template: ` <sebng-chip id="id" className="className" (onClose)="onClose($event)">Test</sebng-chip> `,
```

lib/src/chip/chip.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Component, Input, ViewEncapsulation, EventEmitter, Output } from "@angular/core";
+
+@Component({
+    selector: "sebng-chip",
+    styleUrls: ["./chip.component.scss"],
+    templateUrl: "./chip.component.html",
+    encapsulation: ViewEncapsulation.None,
```
mario-subo, line 7: we don't do this anymore, just omit encapsulation altogether

lib/src/chip/chip.component.html
DanSallau: Don't see the need for this because we are rendering just one content.

lib/src/chip/chip.component.ts
DanSallau: Why though ?

DanSallau: Not providing one means it takes the default being `Emulated` | `None` . I believe there was a reason why we agreed on the none long ago. When was this changed ?

mario-subo: The issues we faced then were due to the way angular was encapsulating before. I don't think this issue is there anymore. @omarVodiaK do you remeber?

DanSallau: > The issues we faced then were due to the way angular was encapsulating before. I don't think this issue is there anymore. @omarVodiaK do you remeber?

I remember the encapsulation issue and we arrived at that after realising how it messes with our components and styling encapsulations. My question is how was this issue resolved ?

lib/src/chip/chip.component.html
mario-subo: Exactly so you don't need an unecessary wrapper with class content which does nothing. Please use the example I gave you to refactor as this is the angular way and other components are following this logic as well.

lib/src/chip/chip.component.ts
mario-subo: Ok forget about why we used it before. 

https://dev.to/monicafidalgo/what-is-viewencapsulation-in-angular-470o

ViewEncapsulation.Emulated (DEFAULT)
The Emulated mode is the default one. This allows that styles from main HTML propagate to the component but styles defined in this component's @Component decorator are scoped to this component only.

ViewEncapsulation.None
In the None mode, styles from the component propagate back to the main HTML and therefore are visible to all components on the page. Be careful with apps that have None components in the application.

Which one do you think makes more sense to use?

lib/src/chip/chip.component.html
yousifalraheem: so, you mean something like this:
```suggestion
    <ng-content></ng-content>
```

lib/src/chip/chip.component.ts
yousifalraheem: The `None` version makes more sense. We are trying to make as close as possible to an HTML only layout. The emulated version will change class names and add weird stuff just to implement encapsulation for those developers who don't know how to achieve it with CSS, but we do know how to do it, so there is no need for it.

lib/src/chip/chip.component.html
DanSallau: I am aware of the ng-content select Mario and i don't see the need for it here. You use select when you have multiple ng-content, which we don't in this component. It's just a simple chip component and regardless the user can have just one-content. So why the hassle of adding a select/target.

lib/src/chip/chip.component.ts
DanSallau: I am aware of the Encapsulations types and thanks to the issues we had long ago, we decided to use the `None` for our component. We had a reason back then and my question is has the reason changed and why ?

lib/src/chip/chip.component.html
mario-subo: Then you can at least omit the unnecessary wrapper

lib/src/chip/chip.component.ts
mario-subo: If you read what I posted you will see that None enables your component CSS to propagate back to the main html which can cause bugs. 

Do you remember what the issue was originally and why the decided to use None? I prefer to go with default angular behaviour whenever possible.

PR #266, opened 2020-05-14
feat(accordion): new accordion component

DanSallau:
accordion


projects/ng-components-docs/src/app/examples/components/accordion/accordion.component.ts
```diff
@@ -0,0 +1,76 @@
...
+            header: "Accordion List Item 3",
+            content: `
+            <div>
+                <p className="m-0">Ut nemo corporis inventore neque qui. Est quos facere et id praesentium ut in iusto qui. Labore vel est ab.</p>
+                <a href="https://seb.se" target="_blank">
+                    Link to seb.se
+                </a>
+            </div>`,
+        },
+    ];
```
yousifalraheem, line 75: You can utilize the package `lorem-ipsum` to generate text like this and clean up the UI
given you have this:

```typescript
import { loremIpsum } from "lorem-ipsum";
```

You can do this:

```typescript
accordionList: Array<AccordionListItem> = Array(3).fill({
    header: loremIpsum({ units: "words", count: 3 }),
    subHeaderText: loremIpsum({ units: "words", count: 3 }),
    content: {
        title: loremIpsum({ units: "words", count: 3 }),
        desc: loremIpsum({ units: "paragraph" })
    },
});
```

lib/src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        component.id = id;
+        component.className = className;
+
+        fixture.detectChanges();
+        expect(fixture.debugElement.query(By.css(`.${className}`))).toBeTruthy();
+        expect(fixture.debugElement.query(By.css(`#${id}`))).toBeTruthy();
+    });
+
+    it("Should render subheader is included in props", () => {
+        const newAccordionList = JSON.parse(JSON.stringify(accordionList));
```
yousifalraheem, line 85:
```suggestion
        const newAccordionList: Array<AccrodionListItem> = JSON.parse(JSON.stringify(accordionList));
```

yousifalraheem: Also use [deepCopy](https://github.com/sebgroup/frontend-tools/wiki/deepCopy) from our frontend-tools instead of JSON manipulation

lib/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,175 @@
...
+        this.accordionClassName = cn;
+    }
+
+    /** Constructs the `className` to be used in accordion items */
+    constructItemClassName(): void {
+        let cn: string = "accordion-item";
+        cn += " " + (this.iconPosition ? this.iconPosition : "left");
+        cn += " " + (this.iconRotation ? this.iconRotation : "deg-180");
+        cn += this.customIconExpanded ? " transform" : "";
+        this.itemClassName = cn;
```
yousifalraheem, line 86: instead of this, I believe you can construct an object for `ngClass` to consume. @omarVodiaK can back me up on this

lib/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,175 @@
...
+                // Another section is already expanded
+                this.expandOrCollapseSection(this.active);
+            }
+            this.expandOrCollapseSection(index);
+        }
+        this.active = this.active === index ? null : index;
+    }
+
+    contentIsOfTypeArray(content: AccordionContentType): boolean {
+        return content instanceof Array;
```
yousifalraheem, line 112: I believe this is more accurate, you can read more about the difference [here](https://stackoverflow.com/questions/22289727/difference-between-using-array-isarray-and-instanceof-array) 
```suggestion
        return Array.isArray(content);
```

lib/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,175 @@
...
+        this.constructItemClassName();
+        this.heightList = Array(this.list?.length).fill(0);
+    }
+
+    ngAfterViewInit() {
+        this.toggle(this.activeIndex);
+        this.accordionItemRefs.changes.subscribe(r => {
+            setTimeout(() => {
+                this.toggle(this.activeIndex);
+            }, 0);
```
yousifalraheem, line 150: is this necessary?

lib/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,175 @@
...
+    ngOnInit() {
+        this.constructIds();
+        this.constructClassName();
+        this.constructItemClassName();
+        this.heightList = Array(this.list?.length).fill(0);
+    }
+
+    ngAfterViewInit() {
+        this.toggle(this.activeIndex);
+        this.accordionItemRefs.changes.subscribe(r => {
```
yousifalraheem, line 147: if you're not using `r`, I believe you can omit it
```suggestion
        this.accordionItemRefs.changes.subscribe(() => {
```

lib/src/accordion/accordion.pipe.ts
```diff
@@ -0,0 +1,10 @@
+import { DomSanitizer } from "@angular/platform-browser";
+import { Pipe, PipeTransform } from "@angular/core";
+
+@Pipe({ name: "safeHtml" })
+export class SafeHtmlPipe implements PipeTransform {
+    constructor(private sanitizer: DomSanitizer) {}
+    transform(value: string) {
```
yousifalraheem, line 7: missing return type

lib/src/accordion/accordion.component.ts
DanSallau: I can actually use angular ngClass too, just got caught up with the react implementation i guess.

lib/src/accordion/accordion.component.ts
DanSallau: isArray support wasn't that good before. Unless if its updated, i tend to avoid it.

lib/src/accordion/accordion-style.pipe.ts
```diff
@@ -0,0 +1,12 @@
+import { DomSanitizer } from "@angular/platform-browser";
+import { Pipe, PipeTransform } from "@angular/core";
+
+@Pipe({ name: "dynamicStyle" })
+export class DynamicStylePipe implements PipeTransform {
+    constructor(private sanitizer: DomSanitizer) {}
```
mario-subo, line 6: where are you using the sanitizer? do you even need it in this file?

mario-subo:
sorry I approved by mistake haha

lib/src/accordion/accordion.component.ts
DanSallau: Alternatively you can use ```this.changeDetector.detectChanges();```. Since i have that on every viewChecked lifecycle. We can do away with the timeout

lib/src/accordion/accordion.component.spec.ts
DanSallau: Is that really necessary ? It's angular and unit test at that .

lib/src/accordion/accordion.component.ts
DanSallau: Here is what it will look like if i am to go with `ngClass` . `[ngClass]="[className, { 'alternative-accordion': alternative }]"` . This however has some issue. Since both className and alternative are optional. It doesn't work and often times result in a comma after the first prop when not provided. I think having the string `itemClassName` should suffice here.

lib/src/accordion/accordion-style.pipe.ts
DanSallau: We are not. It slipped through accidentally.

DanSallau:
> sorry I approved by mistake haha

Still count. 
image

lib/src/accordion/accordion.component.spec.ts
yousifalraheem: yep, it's necessary to catch any changes we make on the APIs

projects/ng-components-docs/src/app/examples/components/accordion/accordion.component.ts
```diff
@@ -0,0 +1,55 @@
...
+            </g>
+        </svg>
+    `;
+    angleDoubleRightIcon: string = `
+        <svg name="angle-double-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512">
+            <path d="M166.9 264.5l-117.8 116c-4.7 4.7-12.3 4.7-17 0l-7.1-7.1c-4.7-4.7-4.7-12.3 0-17L127.3 256 25.1 155.6c-4.7-4.7-4.7-12.3 0-17l7.1-7.1c4.7-4.7 12.3-4.7 17 0l117.8 116c4.6 4.7 4.6 12.3-.1 17zm128-17l-117.8-116c-4.7-4.7-12.3-4.7-17 0l-7.1 7.1c-4.7 4.7-4.7 12.3 0 17L255.3 256 153.1 356.4c-4.7 4.7-4.7 12.3 0 17l7.1 7.1c4.7 4.7 12.3 4.7 17 0l117.8-116c4.6-4.7 4.6-12.3-.1-17z" />
+        </svg>
+    `;
+
+    accordionList: Array<AccrodionListItem> = [...Array(3).keys()].map((i: number) => {
```
yousifalraheem, line 38: Optional: But I think it works without the `keys`
```suggestion
    accordionList: Array<AccrodionListItem> = [...Array(3)].map((i: number) => {
```

DanSallau: Yah but i will be undefine unless i use the index. I am using the i value though.

projects/ng-components-docs/src/app/examples/components/accordion/accordion.component.ts
```diff
@@ -0,0 +1,55 @@
...
+        return {
+            header: loremIpsum({ units: "words", count: 3 }),
+            subHeaderText: loremIpsum({ units: "words", count: 3 }),
+            content:
+                i === 2
+                    ? `<div>
+        <p className="m-0">Ut nemo corporis inventore neque qui. Est quos facere et id praesentium ut in iusto qui. Labore vel est ab.</p>
+        <a href="https://seb.se" target="_blank">
+            Link to seb.se
+        </a`
```
yousifalraheem, line 48: You are missing the last parts of the `anchor` tag and a closing `div`. You can also generate HTML with `lorem-ipsum`
```suggestion
                    ? `<div>${loremIpsum({ units: "paragraph", format: "html" })}<a href="https://seb.se" target="_blank">Link to seb.se</a></div>`
```

package-lock.json
```diff
@@ -19317,7 +19332,7 @@
         "rfdc": {
             "version": "1.1.4",
             "resolved": "https://registry.npmjs.org/rfdc/-/rfdc-1.1.4.tgz",
-            "integrity": "sha512-5C9HXdzK8EAqN7JDif30jqsBzavB7wLpaubisuQIGHWf2gUXSpzy6ArX/+Da8RjFpagWsCn+pIgxTMAmKw9Zug==",
+            "integrity": "sha1-unLME2egzNnPgahws7WL060H+MI=",
```
yousifalraheem, line 19335: We have the same issue here

yousifalraheem: Make sure you resolve these integrity issues

DanSallau: resolved

lib/src/accordion/accordion.component.ts
yousifalraheem: can you move this function and the one before `constructClassName` to an `ngClass`?

DanSallau: No, I couldn't because of the issue i explain above. I think constructing a string variable `itemClassName` is the right solution considering `className` and `alternative` are both optional.

yousifalraheem: Can you at least return an ngClass object instead, I don't want us to deal with class name spaces manually
```suggestion
        return {
            "accordion-item", true,
            [this.iconPosition || "left"]: true,
            [this.iconRotation || "deg-180"]: true,
            "transform": this.customIconExpanded,
            [this.itemClassName]: this.itemClassName
        };
```

DanSallau: Right now you have a structured variable within angular lifecycle that is only recalculated when necessary or one of the the props changes. Returning a function here is not the angular way and bad practice. It simply means getting recalculated when ever the ui is re-rendered or a prop change regardless. If we have to, then alternatively we can use pipe and have something like `[ngClass]=" src | calculateClass :  [iconPosition, iconRotation, customIconExpanded] "` which is the angular way. Personally i don't find that necessary.

PR #274, opened 2020-05-20
feat(notification): new notification component

DanSallau:
notifications


lib/src/notification/notification.component.scss
```diff
@@ -0,0 +1,286 @@
...
+                fill: $close-hover;
+            }
+        }
+    }
+}
+
+.custom-notification {
+    position: fixed;
+    display: block;
+    z-index: 9999;
```
yousifalraheem, line 42: can you use the `z-index` variable from bootstrap? I think there is one defined for `toast` component or something similar, if not, just use the highest one which I think is the `tooltip`

lib/src/notification/notification.component.scss
```diff
@@ -0,0 +1,286 @@
...
+            fill: $white;
+            position: relative;
+            display: block;
+            width: 16px;
+            height: 16px;
+        }
+    }
+
+    &.style-slide-in {
+        width: 300px;
```
yousifalraheem, line 89: move this to a variable

lib/src/notification/notification.component.scss
```diff
@@ -0,0 +1,286 @@
...
+            transform: translateX(-350px);
+            left: 20px;
+            top: 15px;
+        }
+
+        &.top-right {
+            transform: translateX(350px);
+            right: 20px;
+            top: 15px;
+        }
```
yousifalraheem, line 113: If you define the width as a variable then we can calculate the translate value:
```scss
// This
transform: translateX($notificationWidth + 50);
// And this
transform: translateX(-($notificationWidth + 50));
```

lib/src/notification/notification.component.scss
```diff
@@ -0,0 +1,286 @@
...
+
+        .content-wrapper {
+            border-radius: 4px;
+            will-change: box-shadow, top, left;
+            transition: box-shadow 80ms linear, top 100ms ease, left 100ms ease, background-color 200ms ease;
+
+            &.clickable {
+                &:active {
+                    box-shadow: 0 0 0 0 rgba(0, 0, 0, 0), 1px 2px 11px rgba(0, 0, 0, 0.5);
```
yousifalraheem, line 127: move this to a variable

lib/src/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+            component.dismissTimeout = 500;
+            component.onDismiss = () => (component.toggle = false);
+
+            fixture.detectChanges();
+
+            expect(fixture.debugElement.query(By.css(".open"))).toBeTruthy();
+            await fixture.whenStable();
+            await setTimeout(() => {
+                expect(fixture.debugElement.query(By.css(".open"))).toBeTruthy();
+            }, 600);
```
yousifalraheem, line 111: instead of testing this asynchronously which might cause issues, you can mock the timers and synchronously travel through time by using fake times. You can read more about this [here](https://jasmine.github.io/2.0/introduction.html#section-Mocking_the_JavaScript_Timeout_Functions)

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
+import { OnChanges, Component, ViewEncapsulation, Input, Output, EventEmitter, OnInit, SimpleChanges, OnDestroy } from "@angular/core";
+
+export type NotificationStyle = "slide-in" | "bar";
+export type NotificationPosition = "bottom-left" | "bottom-right" | "top-left" | "top-right" | "top" | "bottom";
+export type NotificationTheme = "purple" | "primary" | "danger" | "success" | "warning" | "inverted";
+
+export interface NotificationAction {
+    text: string;
+    action: () => void;
```
yousifalraheem, line 9: This is cleaner
```suggestion
    action: VoidFunction;
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+    public showNotificationTitle: boolean = false;
+    private timerRef: number = null;
+    private defaultTimeout: number = 5000;
+
+    // helper functions
+
+    /**
+     * Get the style class based on the theme passed through the props
+     * @param {string} style The style passed through the props
+     * @returns {void} The style class
```
yousifalraheem, line 46: You don't need this line
```suggestion
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+     */
+    private getStyleClass(): void {
+        let styleClass: string = "style-";
+        if (this.style && ["slide-in", "bar"].indexOf(this.style) !== -1) {
```
yousifalraheem, line 50:
```suggestion
        if (["slide-in", "bar"].some((style) => style == this.style) {
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+            styleClass += this.style;
+        } else {
+            styleClass += "slide-in";
+        }
+
+        this.notificationClassNames += " " + styleClass;
+
+        this.showNotificationTitle = this.title && this.style === ("style-slide-in" as NotificationStyle);
```
yousifalraheem, line 58: wouldn't this always returns `false` since `this.style` will not have `style-` prefix defined by the user?

Maybe this is what it should look like:
```suggestion
        this.showNotificationTitle = this.title && styleClass === "style-slide-in";
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+        this.showNotificationBody =
+            styleClass === ("style-slide-in" as NotificationStyle) && this.actions && this.actions.length && this.actions.length < 3;
```
yousifalraheem, line 60: You defined `styleClass` as string, so no need for parsing
```suggestion
            styleClass === "style-slide-in" && this.actions?.length < 3;
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Get the theme class based on the theme passed though the props
+     * @returns {void} The theme class
+     */
+    private getThemeClass(): void {
+        let themeClass: string = "theme-";
+        if (this.theme && ["purple", "primary", "danger", "success", "warning", "inverted"].indexOf(this.theme) !== -1) {
```
yousifalraheem, line 69:
```suggestion
        if (["purple", "primary", "danger", "success", "warning", "inverted"].some((theme) => theme === this.theme)) {
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Get the position class based on the position and style passed through the props
+     * @returns {void} The position class
+     */
+    private getPositionClass(): void {
+        let positionClass: string;
+        const position: string = this.position;
+        if (this.style && ["slide-in", "bar"].indexOf(this.style) !== -1) {
```
yousifalraheem, line 84:
```suggestion
        if (["slide-in", "bar"].some((style) => style === this.style)) {
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+            switch (this.style) {
+                case "slide-in":
+                    if (position && ["bottom-left", "bottom-right", "top-left", "top-right"].indexOf(position) !== -1) {
```
yousifalraheem, line 87:
```suggestion
                    if (["bottom-left", "bottom-right", "top-left", "top-right"].some((pos) => pos === position)) {
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+                        positionClass = position;
+                    } else {
+                        positionClass = "bottom-left";
+                    }
+                    break;
+                case "bar":
+                    if (position && ["top", "bottom"].indexOf(position) !== -1) {
```
yousifalraheem, line 94:
```suggestion
                    if (["top", "bottom"].some((pos) => pos === position)) {
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+                        positionClass = position;
+                    } else {
+                        positionClass = "top";
+                    }
+                    break;
+            }
+        } else {
+            // Should default back to `slide-in`
+            if (position && ["bottom-left", "bottom-right", "top-left", "top-right"].indexOf(position) !== -1) {
```
yousifalraheem, line 103:
```suggestion
            if (["bottom-left", "bottom-right", "top-left", "top-right"].some((pos) => pos === position)) {
```

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,177 @@
...
+            if (this.toggle && !this.persist) {
+                this.startTimer();
+            } else {
+                this.clearTimer();
+            }
+
+            this.setClassNames();
+        }
+
+        if (changes.style || changes.position || changes.theme || changes.className || changes.actions) {
```
yousifalraheem, line 162: Wouldn't this be easier to maintain
```suggestion
        if (["style", "position", "theme", "className", "actions"].some((prop) => changes.hasOwnProperty(prop))) {
```

projects/ng-components-docs/src/app/examples/components/notification/notification.component.html
```diff
@@ -0,0 +1,53 @@
...
+            [toggle]="notification2Toggle"
+            title="Notification Title"
+            [message]="description"
+            [actions]="notificationActions"
+            (onDismiss)="onDismiss()"
+            persist="true"
+            position="bottom-right"
+            theme="primary"
+        ></sebng-notification>
+        <button sebng-btn (click)="onToggle2(!notification1Toggle)">With Actions</button>
```
yousifalraheem, line 29:
```suggestion
        <button sebng-btn (click)="onToggle2(!notification2Toggle)">With Actions</button>
```

projects/ng-components-docs/src/app/examples/components/notification/notification.component.html
```diff
@@ -0,0 +1,53 @@
...
+            position="bottom"
+            theme="danger"
+        >
+            <div className="notification-message">
+                <button *ngIf="description" class="btn btn-sm btn-secondary text-light px-1 py-0" (onClick)="(null)">
+                    Click here
+                </button>
+            </div>
+        </sebng-notification>
+        <button sebng-btn (click)="onToggle3(!notification1Toggle)">Bar</button>
```
yousifalraheem, line 51:
```suggestion
        <button sebng-btn (click)="onToggle3(!notification3Toggle)">Bar</button>
```

projects/ng-components-docs/src/app/examples/components/notification/notification.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    public description: string = "Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.";
+    public notificationActions: Array<NotificationAction> = [
+        { text: "Yes, I'm in", action: () => this.onToggle2(false) },
+        { text: "Ignore", action: () => this.onToggle2(false) },
+    ];
+
+    onClick(e: MouseEvent) {
+        console.log("Notification clicked");
+    }
```
yousifalraheem, line 21: is this necessary? No one will check the console while viewing the documentation

lib/src/notification/notification.component.ts
DanSallau: I get what you mean , but i disagree with the approach. `style` , `position` and `theme` here are props in the component , not some dynamic array property and your approach above risk sending that message. Why should i utilised an array `some` here which is another callback and implementation on it own just to check on props that are predefined 
?

lib/src/notification/notification.component.scss
DanSallau: While the styling changes might make sense, it would represent a sharp decline from it resemblance with our current react notification component stylings. If they are necessary then perhaps we should be moving the same to react component too. The styling file here is exact replica of what we have in react component.

lib/src/notification/notification.component.spec.ts
DanSallau: The link above is a different scenario where the test object to be spied on is within timeout. It however reminded me of a more feasible solution [fakeAsync](https://angular.io/api/core/testing/fakeAsync)

lib/src/notification/notification.component.ts
DanSallau: Not necessary .

lib/src/notification/notification.component.ts
yousifalraheem: array some is only used to get you a boolean value, the approach is to improve maintainability when trying to add or remove props from the condition

lib/src/notification/notification.component.ts
yousifalraheem: I've recently been using array some for these kinds of conditions because that's what it's meant for. `indexOf` is meant to return the index of the item to be used somehow, while `every` and `some` is meant to verify the existence of all or some of the elements. It also returns a boolean which is easier to work within this condition.

lib/src/notification/notification.component.scss
yousifalraheem: I will be moving it to a variable as soon as I get to it during my refactoring

lib/src/notification/notification.component.ts
DanSallau: I know and i understand yousif. But to me is like you are adding unnecessary step here. The props are objects and angular `ngChanges` handle it cleanly in a way that only the props that change is available in the changes. I will see the need for adding the logic here if we have a lot of props. Think about this too, alternatively i could use a `getter` and `setter` as input which will negate the need for the `ngOnChange` lifecycle. If i have to add another array logic within the onChange, why not eliminate it entirely considering we can.

lib/src/notification/notification.component.ts
DanSallau: Yah this is crucial, the unit test ought have failed too.

lib/src/notification/notification.component.scss
```diff
@@ -0,0 +1,288 @@
+@import "~@sebgroup/bootstrap/scss/variables";
+
+$small: 12px;
```
omarVodiaK, line 3: the smaller font-size that SEB uses is 0.875rem = 14px, i dont think we should come up with our own font size

lib/src/notification/notification.component.scss
```diff
@@ -0,0 +1,288 @@
...
+$notification-width: 300px;
```
omarVodiaK, line 4: remove this and use $toast-width its already defined in the file you imported

lib/src/notification/notification.component.scss
```diff
@@ -0,0 +1,288 @@
+@import "~@sebgroup/bootstrap/scss/variables";
```
omarVodiaK, line 1: There is a lot of custom css in here, the point of having SEBbootstrap is to avoid that.
This https://sebgroup.github.io/bootstrap/toast-notifications contains what you need 
you will notice that the classnames start with **toast** please use them and remove the css

lib/src/notification/notification.component.html
```diff
@@ -0,0 +1,25 @@
+<div class="{{ notificationClassNames }}">
+    <div class="content-wrapper" [class.clickable]="onClick" (click)="handleClick($event)">
```
omarVodiaK, line 2:
```suggestion
    <div class="toast-container" [class.clickable]="onClick" (click)="handleClick($event)">
```

lib/src/notification/notification.component.html
```diff
@@ -0,0 +1,25 @@
...
+        <div class="notification-title" *ngIf="showNotificationTitle">{{ title }}</div>
```
omarVodiaK, line 3:
```suggestion
        <div class="toast-title" *ngIf="showNotificationTitle">{{ title }}</div>
```

lib/src/notification/notification.component.html
```diff
@@ -0,0 +1,25 @@
...
+        <div class="notification-message" *ngIf="message">{{ message }}</div>
```
omarVodiaK, line 4:
```suggestion
        <div class="toast-message" *ngIf="message">{{ message }}</div>
```

lib/src/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+    beforeEach(() => {
+        fixture = TestBed.createComponent(NotificationTestComponent);
+        component = fixture.componentInstance;
+
+        component.toggle = true;
+        onDismiss = spyOn(component, "onDismiss");
+        onClick = spyOn(component, "onClick");
+        component.message = "Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.";
+        component.actions = [
+            { text: "Yes, I'm in", action: () => this.onToggle2(false) },
```
omarVodiaK, line 74: where is onToggle2 declared, we should not need 2 at the end of the function name

lib/src/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+            fixture.detectChanges();
+
+            expect(fixture.debugElement.query(By.css(".style-bar"))).toBeTruthy();
+            expect(fixture.debugElement.query(By.css(".top"))).toBeTruthy();
+            expect(fixture.debugElement.query(By.css(".theme-purple"))).toBeTruthy();
+        });
+
+        it(
+            "Should dismiss when the default timer is done",
+            <any>fakeAsync((): void => {
```
omarVodiaK, line 102: remove any

lib/src/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+
+                expect(fixture.debugElement.query(By.css(".open"))).toBeTruthy();
+
+                tick(600);
+                expect(fixture.debugElement.query(By.css(".open"))).toBeTruthy();
+            })
+        );
+
+        it("Should render with defaults if style or position props passed is not supported", () => {
+            component.style = "bingo" as NotificationStyle;
```
omarVodiaK, line 117: bingo ?

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,176 @@
...
+    selector: "sebng-notification",
+    templateUrl: "./notification.component.html",
+    styleUrls: ["./notification.component.scss"],
+    encapsulation: ViewEncapsulation.None,
+})
+export class NotificationComponent implements OnChanges, OnInit, OnDestroy {
+    @Input() className?: string;
+    @Input() dismissable?: boolean;
+    @Input() dismissTimeout?: number;
+    @Input() message?: string;
```
omarVodiaK, line 22: the message should not be a string, its too straight forward, notifications can contains links or phone number or an email to click on, you should give to the user the possibility to put whatever he wants

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,176 @@
...
+                        positionClass = "bottom-left";
+                    }
+                    break;
+                case "bar":
+                    if (position && ["top", "bottom"].some((p: string) => p === position)) {
+                        positionClass = position;
+                    } else {
+                        positionClass = "top";
+                    }
+                    break;
```
omarVodiaK, line 98: missing default cases here

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    private setClassNames(): void {
+        this.notificationClassNames = "custom-notification";
+        this.getStyleClass();
+        this.getThemeClass();
+        this.getPositionClass();
+
+        this.notificationClassNames += this.toggle ? " open" : "";
```
omarVodiaK, line 140: This is not the way of constructing a class object in angular use this directive https://angular.io/api/common/NgClass

lib/src/notification/notification.component.ts
```diff
@@ -0,0 +1,176 @@
...
+        this.notificationClassNames += this.className ? ` ${this.className}` : "";
+    }
+
+    // events
+
+    handleClick(e: MouseEvent) {
+        this.onClick.emit(e);
+    }
+
+    ngOnChanges(changes: SimpleChanges) {
```
omarVodiaK, line 150: SimpleChanges have those properties   **previousValue: any** and  **currentValue: any** and then inside you check for toggle

projects/ng-components-docs/src/app/examples/components/notification/notification.component.ts
```diff
@@ -0,0 +1,40 @@
...
+        this.notification1Toggle = value;
+    }
+
+    onToggle2(value: boolean): void {
+        this.notification2Toggle = value;
+    }
+
+    onToggle3(value: boolean): void {
+        this.notification3Toggle = value;
+    }
```
omarVodiaK, line 37: one function should be enough

lib/src/notification/notification.component.scss
DanSallau: This is a notification component Omar and solely based on what we already have in react-component. Although i reckoned toast component is a replacement for what we already have, i think we can create another component called toast, tailored based on the bootstrap toast.Notification component along with the loader component is one of the highly adopted component to my understanding and changing it right away might not be a good idea. Perhaps another component is the way to go.

yousifalraheem: Yeah, Toast component is not a replacement for Notification, it's a different one, it even have a different use case.

PR #280, opened 2020-05-22
feat(breadcrumb): new breadcrumb component

DanSallau:


lib/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,25 @@
...
+    selector: "sebng-breadcrumb",
+    templateUrl: "./breadcrumb.component.html",
+    styleUrls: ["./breadcrumb.component.scss"],
+    encapsulation: ViewEncapsulation.None,
+})
+export class BreadcrumbComponent {
+    @Input() className?: string;
+    @Input() id?: string;
+    @Input() list: Array<string>;
+    @Output() onClick = new EventEmitter<ClickProps>();
```
yousifalraheem, line 18:
```suggestion
    @Output() onClick: EventEmitter<ClickProps> = new EventEmitter<ClickProps>();
```

lib/src/breadcrumb/breadcrumb.pipe.ts
```diff
@@ -0,0 +1,11 @@
+import { DomSanitizer } from "@angular/platform-browser";
+import { Pipe, PipeTransform } from "@angular/core";
+
+@Pipe({ name: "safeHtml" })
+export class SafeHtmlPipe implements PipeTransform {
+    constructor(private sanitizer: DomSanitizer) {}
+    transform(value: string) {
+        console.log("Value ", value);
```
yousifalraheem, line 8: remove this
```suggestion
```

lib/src/breadcrumb/breadcrumb.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="custom-breadcrumb" [ngClass]="className" [attr.id]="id">
+    <nav aria-label="breadcrumb">
+        <ol class="breadcrumb">
+            <ng-container *ngFor="let breadcrumb of list; index as i">
+                <li
+                    class="breadcrumb-item"
+                    [ngClass]="{ active: i === list.length - 1 }"
```
mario-subo, line 7: since there is only one condition you could just do `[class.active]="i === list.length-1"`

lib/src/breadcrumb/breadcrumb.component.html
```diff
@@ -0,0 +1,14 @@
...
+                    (click)="handleClick(i, $event)"
+                    [innerHTML]="breadcrumb | safeHtml"
```
mario-subo, line 9: just wondering why the need for the innerHTML using safeHtml here? In case someone wants to pass custom html instead of just label?

DanSallau: Yah the user can pass either .

PR #284, opened 2020-05-27
feat(video): a new video component

DanSallau:


lib/src/tabs/tabs.component.spec.ts
```diff
@@ -70,7 +70,6 @@ describe("TabsComponent", () => {
     });
 
     it("Should fire click event if passed", () => {
-        console.log("Other countries ", fixture.debugElement.queryAll(By.css(".nav-link"))[1]);
```
omarVodiaK, line 73: the exact change is in the notification PR

PR #289, opened 2020-05-29
Progress bar

DanSallau:
Preview

progress1


lib/src/progressBar/progressTheme.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+            case "danger":
+                return "bg-danger";
+            case "warning":
+                return "bg-warning";
+            case "info":
+                return "bg-info";
+            default:
+                return "bg-success";
+        }
+    }
```
mario-subo, line 17: why not just `return "bg-" + value`  ?

lib/src/progressBar/progressTheme.pipe.ts
```diff
@@ -0,0 +1,18 @@
+import { Pipe, PipeTransform } from "@angular/core";
+import { ProggressTheme } from "./progressBar.component";
+
+@Pipe({ name: "proggressTheme" })
```
mario-subo, line 4:
```suggestion
@Pipe({ name: "progressTheme" })
```

lib/src/progressBar/progressTheme.pipe.ts
DanSallau: I think that will defeat the purpose of having  progressTheme type, perhaps we can replace it with string then. Also imagine a user passing an invalid type , since there is no default, then it will match neither of the classes.

lib/src/progressBar/progressBar.component.html
```diff
@@ -0,0 +1,33 @@
+<div class="progress" [ngClass]="className" [attr.id]="id" [ngStyle]="height ? { height: height + 'px' } : null">
```
omarVodiaK, line 1: the only style you want to use here is the height
```suggestion
<div class="progress" [ngClass]="className" [attr.id]="id" [style.height.px]="height ? height : null">
```

projects/ng-components-docs/src/app/examples/components/progressBar/progressBar.component.ts
```diff
@@ -0,0 +1,41 @@
+import { Component, OnInit, OnDestroy } from "@angular/core";
+import { BarItem } from "lib/src/public_api";
```
omarVodiaK, line 2: is it ok to import BarItem from public_api ? i believe we removed this public_api and instead you can import BarItem from the progress component

lib/src/progressBar/progressTheme.pipe.ts
yousifalraheem: I think that would be their fault, and it's easier to capture a mistake when you can see the glitch afterwards. I think you can use `bg-${value}`

DanSallau: There is no mistake here though. The result is a theme , displaying a totally different. I will know right away if i am using the wrong one. Plus regardless whether user pass the right or wrong one, there must always be a theme, a default one at that.

projects/ng-components-docs/src/app/examples/components/progressBar/progressBar.component.ts
DanSallau: The VS code auto import it seems. Yah will reference the component.

lib/src/progressBar/progressBar.component.html
DanSallau: Make sense.

lib/src/progressBar/progressTheme.pipe.ts
yousifalraheem: What if someone wants to use it with a custom class like `bg-nuru`? the switch will not allow that.

DanSallau: Why should it allow it when it doesn't recognise it as a theme and doesn't have a definition for it ? All the themes there are predefined and if you need another, i think you should classNAmes.

lib/src/progressBar/progressBar.component.html
```diff
@@ -0,0 +1,33 @@
...
+            class="progress-bar"
+            *ngFor="let bar of multiBars"
+            [ngClass]="bar.theme | progressTheme"
+            [class.progress-bar-striped]="bar.striped"
+            [class.progress-bar-animated]="bar.animated"
+            role="progressbar"
+            [attr.aria-valuenow]="bar.value"
+            [attr.aria-valuemin]="bar.value"
+            [attr.aria-valuemax]="bar.value"
+            [ngStyle]="{ width: bar.value + '%' }"
```
omarVodiaK, line 13: can you try this
```suggestion
            [style.width.%]="bar.value"
```

lib/src/progressBar/progressBar.component.html
```diff
@@ -0,0 +1,33 @@
...
+        <div
+            class="progress-bar"
+            [ngClass]="theme | progressTheme"
+            [class.progress-bar-striped]="striped"
+            [class.progress-bar-animated]="animated"
+            role="progressbar"
+            [attr.aria-valuenow]="value"
+            [attr.aria-valuemin]="value"
+            [attr.aria-valuemax]="value"
+            [ngStyle]="{ width: value + '%' }"
```
omarVodiaK, line 28:
```suggestion
            [style.width.%]="value"
```

PR #291, opened 2020-06-02
feat(rating): move rating component from angularComp to ng-comp

DanSallau:
Rating is a straight forward component and mostly inconsequential, simply moved from angular components to here. 

Preview https://github.sebank.se/pages/DesignLibrary/AngularComponents/#/rating


lib/src/rating/rating.component.html
```diff
@@ -0,0 +1,48 @@
+<div class="rating-container" [ngClass]="className">
+    <div class="star-rating" [class.readonly]="readOnly" role="radiogroup" aria-labelledby="radioGroup">
```
yousifalraheem, line 2: can you rename `readonly` class to `read-only`

yousifalraheem, line 2: the `aria-labelledby` should be passed by the user if necessary, not hardcoded

lib/src/rating/rating.component.ts
```diff
@@ -0,0 +1,231 @@
...
+    tabCounter: number = 0;
+
+    @ViewChildren("radioItemRefs") radios: QueryList<ElementRef>;
+
+    ngOnInit() {
+        if (this.max) {
+            this.randomIds = Array.apply(null, Array(this.max)).map((item: any) => {
+                const identifier = +Math.floor(Math.random() * 100) + new Date().getTime();
+                return "star-" + identifier;
+            });
```
yousifalraheem, line 53: can you use `randomId` from [frontend-tools](https://github.com/sebgroup/frontend-tools/wiki/randomId)

lib/src/rating/rating.component.ts
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * Provides an array for ngFor to loop through
+     * @returns {Array<number>} An array of ones just to loop through it
+     */
+    getList(): Array<number> {
+        const list: Array<number> = new Array(this.max);
+        for (let i: number = 0; i < list.length; i++) {
+            list[i] = 1;
+        }
+        return list;
```
yousifalraheem, line 66:
```suggestion
        return Array(this.max).fill(1);
```

lib/src/rating/rating.component.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    setTabIndex(i: number): string {
+        if (i === 0) {
+            return "0";
+        }
+        return Math.floor(this.value) === i + 1 ? "0" : "-1";
```
yousifalraheem, line 73:
```suggestion
        return i === 0 ? "0" : Math.floor(this.value) === i + 1 ? "0" : "-1";
```

lib/src/rating/svgStarHollow.svg.html
```diff
@@ -0,0 +1,31 @@
...
+    </defs>
+    <g>
+        <rect fill="none" id="canvas_background" height="6.61537" width="6.61537" y="-1" x="-1" />
+    </g>
+    <g>
+        <title *ngIf="title">{{title}}</title>
+        <g fill-rule="evenodd" fill="none" id="Page-1">
+            <g id="Icons-main-navigation/Star-unselected/Darkblue">
+                <g id="Icon">
+                    <mask transform="translate(2,2) " id="mask-unselected">
```
yousifalraheem, line 23: I think we should remove all IDs since we are rendering at least 5 instances of either of the SVGs.

mario-subo:
Have you made any changes from Angular Components that need my attention? Did you copy over the unit tests and are they passing?

lib/src/rating/rating.component.html
DanSallau: Moved it to the appropriate span element. It doesn't belong here.

lib/src/rating/rating.component.ts
DanSallau: Not necessary.

yousifalraheem: It's OK, we have polyfills for that, I just don't think the function is necessary. If you don't like it you can even use this:
```javascript
"1".repeat(this.max).split("").map(i => +i)
```

PR #293, opened 2020-06-02
feat(table): table

### LINKED ISSUE #46: Table design (update)
### Component review 2019-10-08
There are two small and important changes on table design:
- Bigger table header font: 16 px, medium, Grey 2 (#333333)
- New sorting icon: Normal = Grey 7 (#B2B3B3), Selected= Dark Blue 2 (#007AC7), Hover = Blue (#41B0EE)
- The sorting icon is FontAwesome solid "chevron-up" and solid "chevron-down", on top of each other, width 10 x height 16 px (together)

image

There is more to come, but these changes are the most urgent. So the rest of what we know about tables stays the same for now.

Thank you!
/Ulrika
Design Management

### LINKED ISSUE #292: Table component
Add a table component which follows the SEB design

mario-subo:
fix #46 fix #292

useful reference links:
https://sebgroup.github.io/bootstrap/tables
http://designlibrary.sebank.se/components/table/#code

TODO: 
- [x] Fix sticky position for fixed table
- [x] Provide more examples

- [x] Fix selectable paginated table, handle selected rows in table service instead?
- [x] Add ability to project custom html content for each cell
- [x] Keep selected rows even when changing sort
- [x] Show example for show/hide columns
- [ ] Add polyfills for IE 🤮 
- [x] Add unit tests

Screens:
image


lib/src/table/table-body/table-body.component.html
```diff
@@ -0,0 +1,7 @@
+<tr *ngFor="let row of rows; let i = index" (click)="handleClickRow({ item: row, index: i })">
+    <td *ngIf="selectable" class="table-checkbox">
+        <input type="checkbox" class="form-control" [checked]="selectedRows.includes(i)" />
+    </td>
+
+    <td seb-table-td *ngFor="let column of headerList" [row]="row[column.tableKeySelector]" [valueType]="column.valueType"></td>
```
omarVodiaK, line 6: do we need headerList inside table-body components

lib/src/table/table-td/table-td.component.html
```diff
@@ -0,0 +1,6 @@
+<span *ngIf="valueType === 'string'">{{ row }}</span>
+<span *ngIf="valueType === 'number'">{{ row }}</span>
+<span *ngIf="valueType === 'datetime'">{{ row | date: "yyyy-MM-dd HH:mm:ss" }}</span>
+<span *ngIf="valueType === 'date'">{{ row | date: "yyyy-MM-dd" }}</span>
+<span *ngIf="!valueType">–</span>
+<span *ngIf="!row">–</span>
```
omarVodiaK, line 6: switch case would be nice to have heere
```suggestion
<ng-container [ngSwitch]="valueType">
  <span *ngSwitchCase="'string'">{{ row }}</span>
  <span *ngSwitchCase="'number'">{{ row }}</span>
  <span *ngSwitchCase="'datetime'">{{ row | date: "yyyy-MM-dd HH:mm:ss" }}</span>
  <span *ngSwitchCase="'date'">{{ row | date: "yyyy-MM-dd" }}</span>
  <span *ngSwitchDefault>–</span>
</ng-container>
```

lib/src/table/table-body/table-body.component.html
mario-subo: I think it's because the information for valueType is coming from the header list.

lib/src/table/table-td/table-td.component.html
mario-subo: that's a good suggestion, thanks :)

lib/src/table/table.component.scss
```diff
@@ -0,0 +1,44 @@
...
+        width: 100%;
+        margin-bottom: 1rem;
+
+        table {
+            margin-bottom: 0px;
+        }
+    }
+
+    table {
+        border-collapse: separate !important; /* Don't collapse */
```
omarVodiaK, line 16: i dont get the relation between dont collapse and separate css value, also having an !important there will make life hard for whoever wants to change that style

lib/src/table/table.component.scss
```diff
@@ -0,0 +1,44 @@
...
+        border-spacing: 0 !important;
```
omarVodiaK, line 17: please avoid using !important

lib/src/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+        return [...wordList].join(" ");
+    }
+    return "";
+}
+
+/**
+ * Capitalizes (makes the first character uppercase) any string
+ * @param value the value to capitalize
+ */
+export function capitalize(value: string): string {
```
omarVodiaK, line 27: please use our awesome tools:
https://github.com/sebgroup/frontend-tools/tree/develop/src/capitalize

lib/src/table/table-header/table-header.component.html
```diff
@@ -0,0 +1,19 @@
+<tr *ngIf="list && list.length">
+    <th *ngIf="selectable" class="table-checkbox" (click)="allSelectedClicked.emit()">
+        <input type="checkbox" class="form-control" [checked]="isAllSelected" />
+    </th>
+
+    <th
```
omarVodiaK, line 6: i would suggest to add a custom className here that way each table header can be styled separately from the rest of the headers

lib/src/table/table-th/table-th.component.html
```diff
@@ -0,0 +1 @@
+<small class="font-weight-bolder" [class.text-muted]="!active">{{ label }}</small>
```
omarVodiaK, line 1: this part a bit complicated to accept, first small tag will define a font-size that is smaller than the base, and font-weight-bolder will define a bolder weight (700 i think) and it will have !important since bootstrap does it like that.

This will create complications for whoever wants to use this components with custom styles.

Most of SEB usecases would align with this, since the header is small and bold but some wouldn't, in this case we should care about the edge cases

lib/src/table/table.models.ts
```diff
@@ -0,0 +1,57 @@
+/** The type of a data the values in a column represent  */
+export type TableHeaderListValueType = "number" | "string" | "date" | "datetime" | "bool";
```
omarVodiaK, line 2: i think another type should be added, many tables in SEB will have a superior sign icon to say that the row is clickable, i think we should care for this, do you think it will be good to have **any** as a possible type for the headers column

lib/src/utils.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Creates a human readable text from camel case
+ * @param {string} value the camel case string (example "caseId")
+ * @param {boolean} shouldCapitalizeEvery OPTIONAL should every subsequest word be capitalized? Defaults to false
+ * @returns the a capitalised human readable string (example "Case Id" or "Case id")
+ */
+export function readableFromCamelCase(value: string, shouldCapitalizeEvery?: boolean): string {
```
omarVodiaK, line 7: You Are Making Complex A Very Easy Thing. You Can Add This In Your CSS.
each table header should be able to accept a className to apply whatever styles individually.
If you want to capitalize then:
`text-transform: capitalize; `
you can also try providing the text that you want to appear directly to the header by modifying the type of each header and adding the property text to it

projects/ng-components-docs/src/app/examples/components/table/table-examples.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    constructor(private tableService: TableService<TableObjectType>) {}
+
+    ngOnInit() {
+        // Register your data with the table service.
+        // (Optional): Pass in your config. The service will return
+        // the headerList and rows to be displayed with the Table component.
+        this.tableService.registerDatasource(this.data);
+
+        // Subscribe to the current table and header list
+        this.tableService.currentTable.subscribe({
```
omarVodiaK, line 36: i found a nice way to unsubscribe from obserables without creating an array checking if it has elements and unsubscribing from it.

```typescript
       import { takeUntil } from 'rxjs/operators';
       const notifier$: Subject<any> = new Subject();
        this.tableService.currentTable.pipe(takeUnitl(notifier$)).subscribe();
        
        ngOnDestroy(): void {
          this.notifier$.next();
        }
```

lib/src/table/table.component.scss
mario-subo: yeah this was left over from before SEB bootstrap had the correct styles. Thanks for spotting this, we don't need it, getting rid of it

lib/src/table/table-th/table-th.component.html
mario-subo: Yes you're right, after checking the designs of my project and SEB Bootstrap I noticed this design is actually coming from my project and is not in Design Library/bootstrap. I will remove the whole thing and just use span with default SEB styles.

lib/src/table/table.models.ts
mario-subo: I'm not sure I follow you. What do you men by superior sign icon? As is it now, this value determines the type of data display within one columns. It will default to string.

lib/src/utils.ts
mario-subo: your suggestion would display this
image

instead of 
image

I there any text-transform feature which would be able to display camelCase as human readable?

projects/ng-components-docs/src/app/examples/components/table/table-examples.component.ts
mario-subo: Still looks a bit hacky but it's better than unsubscribing within array. I just wish angular would have a straightforward way of unsubscribing. Does this work with multiple subscriptions as well? I just use the same `takeUnitl(notifier$)` within every subscription and then `notifier$.next();` on destroy?

omarVodiaK: there is many ways to do it, this one looks really clean and yes you can use the same to unsubscribe from all the subscriptions.
The best way to use observables is though async pipe but this case is only when you wanna subscribe and use the observable in the template, async pipe will handle the unsubscription by default

lib/src/utils.ts
omarVodiaK: i cant remember if there is any, but i doubt it.
The idea here is to avoid writing a script to do this for you instead you should be able to populate the header of the table with values that you choose, here is an example:

The label is the value that will be shown to the user, in this example its a key because each header cell have a translate pipe that will grab the key and translate it to the correct value.

you are not using the translate pipe inside the table component, but atleast you will give a choice for who wants to use it to put the correct labels and usually those ones are already well formated in the CMS without the need to transform them using javascript,

```typescript
[
      {
        label: 'label_typeofsaving',
        classNames: ['col-3', 'sort'],
        tableKeySelector: 'policyTypeName',
        active: true,
        asc: true,
        valueType: 'string'
      },
      {
        label: 'label_owner',
        classNames: ['col-2', 'sort'],
        tableKeySelector: 'policyHolder',
        active: null,
        asc: null,
        valueType: 'string'
      },
      {
        label: 'text_investmentadvice',
        classNames: ['col-2', 'sort'],
        tableKeySelector: 'adviceStatus',
        active: null,
        asc: null,
        valueType: 'number'
      },
      {
        label: 'label_value',
        classNames: ['col-2', 'text-right', 'sort'],
        tableKeySelector: 'insuranceAmount',
        active: null,
        asc: null,
        valueType: 'number'
      },
      {
        label: 'label_totaldevelopment',
        classNames: ['col-2', 'text-right', 'sort', 'total-development'],
        tableKeySelector: 'totalDevelopment',
        active: null,
        asc: null,
        valueType: 'number'
      },
      {
        label: '',
        classNames: ['col-1'],
        tableKeySelector: '',
        active: null,
        asc: null,
        valueType: null
      }
    ]
```

lib/src/table/table.models.ts
omarVodiaK: this is what i meant, the last column of the table is an image, having an arrow means that the row is clickable, it can also be any other icon
image

lib/src/table/table-td/table-td.component.html
omarVodiaK: one thing that i forgot here is the default value doesnt need to have "-" all the time, it can be empty, i dont know why you chose to add that character, i suggest to remove it and keep the span blank, what do you think

lib/src/table/table.models.ts
mario-subo: oh I see. Maybe one of the options can be any custom html? So in the TD switch statement I would inject the innerHTML if the TableHeaderListValueType is custom-html ?

lib/src/utils.ts
mario-subo: There is already an api provided for the user to put custom labels using the tableConfig
Here is the TS definition for it 
https://github.com/sebgroup/ng-components/pull/293/files#diff-6249b4175d128008cd94ca578d959de3R15

It will use the `readableFromCamelCase` by default if you do not provide (override) the label.

mario-subo: this is the logic where the table header is constructed and the label used if provided: 
https://github.com/sebgroup/ng-components/pull/293/files#diff-7540fde1b76eeede1a2842a0a140ffcbR231

projects/ng-components-docs/src/app/examples/components/table/table-examples.component.ts
mario-subo: ok added

lib/src/utils.ts
mario-subo: So since this transformation is only happening at initialization, once for every column name I don't see this as a performance issue so I will keep it for for now.

lib/src/table/table.models.ts
mario-subo: Ok I implemented the custom html so this UI would be possible using that feature

projects/ng-components-docs/src/app/examples/components/table/table.module.ts
```diff
@@ -0,0 +1,57 @@
...
+                    {
+                        path: "table",
+                        component: TableExamplesComponent,
+                        data: {
+                            title: "Table Component",
+                            description: "Additional description for example (optional)",
+                            sources: [
+                                {
+                                    name: "table-examples.component.html",
+                                    src: require("!raw-loader!./table-examples.component.html"),
```
DanSallau, line 29:
```suggestion
                                    src: require("!raw-loader!./table-examples.component.html").default,
```

projects/ng-components-docs/src/app/examples/components/table/table.module.ts
```diff
@@ -0,0 +1,57 @@
...
+                                    lang: "markup",
+                                },
+                                {
+                                    name: "table-examples.component.ts",
+                                    src: require("!raw-loader!./table-examples.component.ts"),
```
DanSallau, line 34:
```suggestion
                                    src: require("!raw-loader!./table-examples.component.ts").default,
```

projects/ng-components-docs/src/app/examples/components/table/table.module.ts
```diff
@@ -0,0 +1,57 @@
...
+                            ],
+                        },
+                    },
+                ],
+            },
+            {
+                path: "api",
+                component: ApiListComponent,
+                data: {
+                    sources: [require("!raw-loader!../../../../../../../lib/src/table/table.component.ts")],
```
DanSallau, line 46:
```suggestion
                    sources: [require("!raw-loader!../../../../../../../lib/src/table/table.component.ts").default],
```

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,388 @@
+import { Injectable, Output } from "@angular/core";
+import { TableHeaderListItem, TableConfig, SortInfo, TableHeaderListValueType } from "./table.models";
+// import { readableFromCamelCase } from "@sebgroup/frontend-tools/dist/dateDiff";
+// import { dateDiff } from "@sebgroup/frontend-tools/dist/dateDiff"; // NOTE: try this if sort table by date is not working with current implementation
```
yousifalraheem, line 4: forgot to remove this?

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    /**
+     * MAKE PAGINATED TABLE
+     * @param table the table you wish to paginate
+     * @param maxItems the number of max items on one page
+     * @returns the paginated table as array or arrays
+     */
+    private makePaginatedTable(table: Array<T>, maxItems: number): Array<Array<T>> {
+        const paginatedTable = [];
```
yousifalraheem, line 217: type?

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * GET PAGINATION OFFSET
+     * Calculates the offset (number of pages displayed) for pagination
+     * @returns the offset as integer
+     */
+    public getPaginationOffset(): number {
+        if (this._paginatedTable) {
+            const PTableLength: number = this._paginatedTable.length;
+            if (PTableLength > 5) {
+                const offset = Math.round(PTableLength / 2);
```
yousifalraheem, line 296:
```suggestion
                const offset: number = Math.round(PTableLength / 2);
```

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,388 @@
...
+                const max = 10;
```
yousifalraheem, line 297:
```suggestion
                const max: number = 10;
```

projects/ng-components-docs/src/app/examples/components/table/table-examples.component.html
```diff
@@ -0,0 +1,46 @@
+<!-- <div class="row">
+    <div class="col">
+        <label>
+            <input type="checkbox" (click)="selectable = !selectable" />
+            Selectable
+        </label>
+    </div>
+</div> -->
```
yousifalraheem, line 8: forgot this?

lib/src/table/table.component.scss
```diff
@@ -0,0 +1,46 @@
...
+            vertical-align: middle;
+        }
+
+        .table-checkbox {
+            width: 3rem;
+            vertical-align: middle;
+            input {
+                width: 1rem;
+                margin: auto;
+                min-height: auto !important;
```
yousifalraheem, line 33: remove `!important` and fix it with specificity

lib/src/table/table.service.ts
yousifalraheem, line 217:
```suggestion
        const paginatedTable: Array<Array<T>> = [];
```

mario-subo: you mean `T[][]` 🤣

lib/src/table/table.component.scss
mario-subo: thanks will do that

projects/ng-components-docs/src/app/examples/components/table/table-examples.component.html
```diff
@@ -0,0 +1,37 @@
...
+    </div>
+
+    <div class="col-12">
+        <label>Table without sorting</label>
+        <seb-table [rows]="data" [headerList]="headerList"></seb-table>
+    </div>
+
+    <div class="col-12">
+        <label>Table with selectable rows</label>
+        <seb-table [rows]="data" [headerList]="headerList" [selectable]="true"></seb-table>
```
kherP, line 30: will it be better if we include the select all example here so when we click on select all checkbox, all rows will be selected?

lib/src/table/table-body/table-body.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component, Input, EventEmitter, Output } from "@angular/core";
+import { TableHeaderListItem, TableRowClickedEvent } from "../table.models";
+
+@Component({
+    selector: "[seb-table-body]",
+    templateUrl: "./table-body.component.html",
+})
+export class TableBodyComponent {
+    @Input() headerList: Array<TableHeaderListItem> = [];
+    @Input() rows: Array<any> = [];
```
kherP, line 10: is it possible to make it generic instead of using `any` here?

lib/src/table/table-body/table-body.component.ts
```diff
@@ -0,0 +1,23 @@
...
+    @Input() selectable?: boolean = false;
+    @Input() selectedRows: number[];
```
kherP, line 12: Array<number> or number[]? haha 😄

lib/src/table/table.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    templateUrl: "./table.component.html",
+    styleUrls: ["./table.component.scss"],
+    encapsulation: ViewEncapsulation.None,
+})
+export class TableComponent {
+    @Input() headerList: Array<TableHeaderListItem> = [];
+    @Input() sortInfo?: SortInfo;
+    @Input() selectable?: boolean = false;
+    @Input() isAllSelected?: boolean = false;
+    @Input() rows: Array<any> = [];
```
kherP, line 15: can we make it generic instead of using `any`?

lib/src/table/table.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    @Input() fixedHeight?: string;
+    @Input() selectedRowIndexes: number[] = [];
```
kherP, line 17: another Array<number> or number[] xD

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,386 @@
...
+import { readableFromCamelCase } from "../utils";
+import { BehaviorSubject } from "rxjs";
+
+@Injectable({
+    providedIn: "root",
+})
+export class TableService<T extends object> {
+    // ------------- INITIALISATION -------------
+    // ============ TABLE =======================
+    private _table: T[];
```
kherP, line 13: haha, I'm not sure if we should standardize the usage of `Array`

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,386 @@
...
+
+        return {};
+    }
+
+    /**
+     * RELOAD TABLE:
+     * Responsible for recalculation (or initialisation) of everything needed for the table including the sort and pagination and rebuilds all the table components
+     */
+    private reloadTable(): void {
+        const table: T[] = this.table && this.table.length ? [...this.table] : [];
```
kherP, line 113: will there be multiple level for the data? if there is, we might need to deep clone the array

lib/src/table/table-header/table-header.component.html
```diff
@@ -0,0 +1,17 @@
+<tr *ngIf="list && list.length">
+    <th *ngIf="selectable" class="table-checkbox" (click)="allSelectedClicked.emit()">
```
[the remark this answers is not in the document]
DanSallau: While surfing below, i realised it comes from table.component.ts which is under your control. You can ignore the above comment if thats the case.

lib/src/table/table-td/table-td.component.html
```diff
@@ -0,0 +1,8 @@
+<ng-container [ngSwitch]="valueType">
+    <span *ngSwitchCase="'string'">{{ row }}</span>
+    <span *ngSwitchCase="'number'">{{ row | number }}</span>
+    <span *ngSwitchCase="'datetime'">{{ row | date: "yyyy-MM-dd HH:mm:ss" }}</span>
+    <span *ngSwitchCase="'date'">{{ row | date: "yyyy-MM-dd" }}</span>
+    <span *ngSwitchCase="'custom-html'" [innerHtml]="row"></span>
```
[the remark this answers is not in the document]
mario-subo: oh yes thank you! will do

lib/src/table/table.component.ts
mario-subo: same here, not sure if we can make this generic since the component willl be used in html using the selector. there is no way to pass the typescript model of the rows like we can in the table service.

projects/ng-components-docs/src/app/examples/components/table/table-examples.component.html
mario-subo: I will refactor this so that this example uses it's own table service. I can only use one service per component

lib/src/table/table.service.ts
mario-subo: > can we also add example for custom header?

@kherP like a header which has custom html? I don't have that feature yet but I can add it if you like.

mario-subo:
> can we also add example for custom header?

@kherP you mean like a header which has custom html? I don't have that feature yet but I can add it if you like.

lib/src/table/table.component.html
```diff
@@ -0,0 +1,21 @@
+<div [style.height]="fixedHeight" class="table-responsive seb-table" [class.seb-table-fixed-height]="!!fixedHeight">
```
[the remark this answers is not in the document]
mario-subo: I want the user to be able to pass `20vmin` or `50rem` `auto` or anything else not just px.

lib/src/table/table-td/table-td.component.html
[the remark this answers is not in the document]
mario-subo: Can we move your 3 safe pipes from each of your components into a common folder please @DanSallau ?

lib/src/table/table.service.ts
mario-subo: for now no multiple levels are supported yet so this should be fine for now

lib/src/table/table.component.ts
mario-subo: I will resolve this and keep the convo going in your original comment here https://github.com/sebgroup/ng-components/pull/293#discussion_r441447600

kherP:
> > can we also add example for custom header?
> 
> @kherP you mean like a header which has custom html? I don't have that feature yet but I can add it if you like.

i was thinking the column header might not necessary be the same as the attribute, but it would be great if we can add custom header, example of use cases:
image

mario-subo:
> > > can we also add example for custom header?
> > 
> > 
> > @kherP you mean like a header which has custom html? I don't have that feature yet but I can add it if you like.
> 
> i was thinking the column header might not necessary be the same as the attribute, but it would be great if we can add custom header, example of use cases:
> image

Ok I see. This is current possible but I did not have an example for it. I've added it now. I'll update the screenshot.

lib/src/table/table-body/table-body.component.ts
mario-subo: @kherP We discussed it and we don't think it's possible. If you know a way let me know. I will resolve this for now.

lib/src/table/safe-html.pipe.ts
```diff
@@ -0,0 +1,14 @@
...
+export class SafeHtmlPipe implements PipeTransform {
+    constructor(private sanitizer: DomSanitizer) {}
+    transform(value: string): SafeHtml {
+        if (value && value.length) {
+            return this.sanitizer.bypassSecurityTrustHtml(value);
+        } else {
+            return "";
+        }
+    }
+}
```
yousifalraheem, line 14: can you place this pipe outside the component folder, similar to what we have with React. Maybe a folder called `__utils`

lib/src/table/table-td/table-td.component.html
```diff
@@ -0,0 +1,8 @@
+<ng-container [ngSwitch]="valueType">
+    <span *ngSwitchCase="'string'">{{ row }}</span>
+    <span *ngSwitchCase="'number'">{{ row | number }}</span>
+    <span *ngSwitchCase="'datetime'">{{ row | date: "yyyy-MM-dd HH:mm:ss" }}</span>
+    <span *ngSwitchCase="'date'">{{ row | date: "yyyy-MM-dd" }}</span>
```
yousifalraheem, line 5: make the date formats something the user can pass with a prop

lib/src/table/table.component.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+            (rowClicked)="handleSelectRow($event)"
+            (selectAllClicked)="handleSelectAll()"
+        ></seb-table>
+    `,
+})
+class TableTestComponent {
+    @ViewChild(TableComponent) tableComponent: TableComponent;
+    rows: TestTableDataType[] = [
+        { label: "Nigeria", value: "NN" },
+        { label: "Iraq", value: "ID" },
```
yousifalraheem, line 33: 😋 
```suggestion
        { label: "Iraq", value: "IQD" },
```

lib/src/table/table.service.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    let pagination: TableConfig["pagination"];
+    let noOfPages: number;
+    let sort: TableConfig<TestTableDataType>["sort"];
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({});
+        service = TestBed.inject<TableService<TestTableDataType>>(TableService);
+        datasource = [
+            { country: "Nigeria", currency: "NN" },
+            { country: "Iraq", currency: "ID" },
```
yousifalraheem, line 23:
```suggestion
            { country: "Iraq", currency: "IQD" },
```

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,436 @@
...
+    /**
+     * INIT CONFIG TYPES
+     * @returns the initial configuration for config types
+     */
+    private initConfigTypes(): TableConfig<T>["types"] {
+        if (this.table && this.table.length) {
+            const types: TableConfig<T>["types"] = this.table.reduce((result, item) => {
+                for (const key in item) {
+                    if (item[key]) {
+                        const type = (typeof item[key]).toLowerCase();
```
yousifalraheem, line 96:
```suggestion
                        const type: string = (typeof item[key]).toLowerCase();
```

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,436 @@
...
+     * @param table the table you wish to paginate with meta data
+     * @param maxItems the number of max items on one page
+     * @returns the paginated table as array or arrays
+     */
+    private makePaginatedTable(table: TableRowWithMeta<T>[], maxItems: number): TableRowWithMeta<T>[][] {
+        const paginatedTable: TableRowWithMeta<T>[][] = [];
+
+        while (table.length > 0) {
+            paginatedTable.push(table.splice(0, maxItems));
+        }
```
yousifalraheem, line 229: I think you are using while here like an `if` statement, I think you should just convert it o an if statement
```suggestion
        if (table.length > 0) {
            paginatedTable.push(table.splice(0, maxItems));
        }
```

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,436 @@
...
+    };
+
+    /**
+     * CHECK IS ALL SELECTED
+     * Checks the selected rows of each table and check returns true/false if all rows in every table are selected
+     * @returns {boolean}
+     */
+    private checkIsAllSelected(): boolean {
+        return this._selectedRows?.map((e, i) => e?.length === this._paginatedTable[i]?.length).every(e => e === true);
+    }
```
yousifalraheem, line 294: If `e` is boolean, you don't need to compare it to `true`
```suggestion
        return this._selectedRows?.map((e, i) => e?.length === this._paginatedTable[i]?.length).every(e => e);
    }
```

lib/src/table/table.service.ts
```diff
@@ -0,0 +1,436 @@
...
+
+    /**
+     * CHECK HAS SOME SELECTIONS
+     * Checks if any rows (at least 1) are selected
+     * @returns {boolean}
+     */
+    private checkHasSelection(): boolean {
+        return this._selectedRows?.map((e, i) => e?.length > 0).some(e => e === true);
```
yousifalraheem, line 302:
```suggestion
        return this._selectedRows?.map((e, i) => e?.length > 0).some(e => e);
```

lib/src/table/table.service.ts
mario-subo: it's mutating the value so the while needs to be there

PR #297, opened 2020-06-03
feat(timeline): new timeline component

DanSallau:


lib/src/timeline/timeline.component.scss
```diff
@@ -0,0 +1,338 @@
+$white: #fff;
+$gray-100: #f8f8f8;
+$gray-300: #dedede;
+$blue: #41b0ee;
+$blue-darker: #007ac7;
```
yousifalraheem, line 5:
```suggestion
@import "~@sebgroup/bootstrap/scss/core";
```

mario-subo:
Have you made any changes from Angular Components that need my attention? Did you copy over the unit tests and are they passing?

PR #349, opened 2020-06-25
Feature/timer

DanSallau:


lib/src/timer/timer.component.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    ngOnChanges(changes: SimpleChanges) {
+        if (changes.duration) {
+            if (changes.duration.currentValue !== null && changes.duration.currentValue !== undefined) {
+                this.setTimerValue();
+            }
+        }
+    }
```
mario-subo, line 59: Since you're only trancing one value you can avoid the on init and on changes and just put the logic inside the setter of duration.

lib/src/timer/timer.component.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    ngAfterContentChecked() {
+        this.changeRef.detectChanges();
```
mario-subo, line 62: can you check if you really need this. I think it should work without it

lib/src/timer/timer.module.ts
```diff
@@ -0,0 +1,10 @@
+import { NgModule } from "@angular/core";
+import { CommonModule } from "@angular/common";
+import { TimerComponent } from "./timer.component";
+
+@NgModule({
+    declarations: [TimerComponent],
+    imports: [CommonModule],
+    exports: [TimerComponent],
+})
+export class TimerlModule {}
```
mario-subo, line 10: typo -> TimerModule

lib/src/timer/timer.component.ts
```diff
@@ -0,0 +1,64 @@
...
+
+@Component({
+    selector: "sebng-timer",
+    templateUrl: "./timer.component.html",
+    encapsulation: ViewEncapsulation.None,
+})
+export class TimerComponent implements OnInit, OnChanges {
+    @Input() id?: string;
+    @Input() className?: string;
+    @Output() callback = new EventEmitter<void>();
```
mario-subo, line 24: please call this something more meaningful like onTimerEnd

DanSallau: Yah, make sense

lib/src/timer/timer.component.ts
DanSallau: YEs we need it .

PR #351, opened 2020-06-28
Feature/loader

DanSallau:
loader


lib/src/loader/loader.component.scss
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+$offset: 200;
+$duration: 1.5s;
+$duration-sq: 1.2s;
+$loader-xs: 1rem;
+$loader-sm: 2.125rem;
+$loader-md: 4.063rem;
+$loader-lg: 6.25rem;
```
mario-subo, line 17: image
Please use the existing SEB Bootstrap variables where possible

lib/src/loader/loader.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="{{ className }}" [ngClass]="size | loaderClasses: cover:fullscreen:type:backdrop" [attr.role]="role">
```
mario-subo, line 1: wouldn't this result in a class named undefined if you don't pass a custom className?

lib/src/loader/loader.component.ts
```diff
@@ -0,0 +1,31 @@
...
+    /** Dims the background the background to indicate UI interactions are blocked */
+    @Input() backdrop?: boolean;
+    /** Screen reader text. Default is `Loading...` */
+    @Input() srText?: string;
+    /** Show or hide the loader. Default is `true` */
+    @Input() toggle?: boolean;
+
+    @Input() role?: string = "status";
+
+    @Input() className?: string;
```
yousifalraheem, line 30:
```suggestion
    /** HTML role value. Default is `status`  */
    @Input() role?: string = "status";
    /** Element class name */
    @Input() className?: string;
```

PR #358, opened 2020-07-01
Feature/datepicker

mario-subo:
Useful links:
https://designlibrary.sebgroup.com/components/date-picker/#code
https://sebgroup.github.io/bootstrap/datepicker
https://ng-bootstrap.github.io/#/components/datepicker/overview
https://sebgroup.github.io/ng-bootstrap/datepicker
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

Design:
image

My TODO: 
- [ ] add month picker
- [ ] add more unit tests

Screens:

Notes:
Should add support for re #157 ?


lib/src/datePicker/safe-html.pipe.ts
```diff
@@ -0,0 +1,10 @@
+import { DomSanitizer } from "@angular/platform-browser";
+import { Pipe, PipeTransform } from "@angular/core";
+
+@Pipe({ name: "safeHtml" })
+export class SafeHtmlPipe implements PipeTransform {
```
kherP, line 5: If not mistaken, there is similar pipe, I wonder if we should create a folder to store the pipes/ functions that can be reused by each component

lib/src/datePicker/date-picker.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@Component({
+    selector: "sebng-date-picker",
+    templateUrl: "./date-picker.component.html",
+    styleUrls: ["./date-picker.component.scss"],
+    providers: [CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR],
+    encapsulation: ViewEncapsulation.None,
+})
+export class DatePickerComponent implements ControlValueAccessor {
+    @Input() placeholder: string = "yyyy-mm-dd";
```
kherP, line 19: is there a way to set the datepicker to only show months and years? Like month picker, year picker

lib/src/datePicker/safe-html.pipe.ts
mario-subo: yeah I moved it to __utils folder and others can reuse it instead of creating their own (By others I mean @DanSallau hahaha)

lib/src/datePicker/date-picker.component.ts
mario-subo: looking into it. Also considering native month and date picker

mario-subo:
As discussed I'm working on the #364 instead of this

PR #408, opened 2020-07-20
Docs/fix docs styles

yousifalraheem:


projects/ng-components/lib/datepicker/datepicker.component.ts
```diff
@@ -188,6 +188,7 @@ export class DatepickerComponent implements ControlValueAccessor {
     }
 
     // get and set accessor----------------------
+    @Input()
```
mario-subo, line 191: This is not a custom input but the value you get with NgModule.

mario-subo: you should remove it.

projects/ng-components/lib/rating/rating.component.ts
```diff
@@ -209,20 +209,21 @@ export class RatingComponent implements ControlValueAccessor, OnInit {
     onChangeCallback: (_: any) => void = () => {};
 
     // get and set accessor----------------------
+    @Input()
```
mario-subo, line 212: same here

projects/ng-components/lib/slider/slider.component.ts
```diff
@@ -270,19 +275,14 @@ export class SliderComponent implements OnInit, OnChanges, AfterContentChecked,
         return this.size / this.step <= maxNumberOfStepsToAllowTransition;
     }
 
-    // controlAccessor
-    private innerValue: number;
-
-    private onTouchedCallback: () => void;
-    private onChangeCallback: (_: any) => void;
-
+    @Input()
```
mario-subo, line 278: same here

projects/ng-components/lib/stepper/stepper.component.ts
```diff
@@ -55,17 +55,18 @@ export class StepperComponent implements ControlValueAccessor {
     }
 
     // get and set accessor----------------------
+    @Input()
```
mario-subo, line 58: ...

projects/ng-components/lib/datepicker/datepicker.component.ts
yousifalraheem: Nope, this is actually a good practice, if you don't include those binding with `[value]` will through an error since Angular will assume that `value` doesn't exist in the component even though it does. Adding this decorator is only to inform Angular that it exists and you can bind to it.

PR #409, opened 2020-07-20
Docs/textarea

mario-subo:
image


projects/docs/src/app/docs-wrapper/pages/textarea-page/textarea-page.component.ts
```diff
@@ -1,24 +1,107 @@
...
+                <sebng-textarea
+                    [placeholder]="extendedFormGroup.value.text.placeholder"
+                    [resizable]="extendedFormGroup.value.toggles.resizable"
+                    [disabled]="extendedFormGroup.value.toggles.disabled"
+                    [readonly]="extendedFormGroup.value.toggles.readonly"
+                    [cols]="extendedFormGroup.value.numbers.cols"
+                    [rows]="extendedFormGroup.value.numbers.rows"
+                    [max]="extendedFormGroup.value.numbers.max"
+                    [(ngModel)]="value"
+                ></sebng-textarea>
```
yousifalraheem, line 19: add label please

mario-subo: I thought we agreed that we are going to slowly remove all labels and let users add their own `<label>Label</label>` on top of the component if they wish so? That's why I didn't provide examples for labels in any components.

yousifalraheem: I don't think I was present in that discussion, because I can argue that it's a tedious process if you want to implement random id with `for` attribute in the label. We should keep the label and showcase it here in the example.

PR #436, opened 2020-07-27
docs: add scully

kherP:
run `npm run build:docs` to build docs with scully


scully.log
```diff
@@ -0,0 +1,222 @@
+No routes defined in ""The option outDir isn't configured, using default folder "[33m.\dist\static\[39m".
+
+No routes defined in ""The option outDir isn't configured, using default folder "[33m.\dist\static\[39m".
```
yousifalraheem, line 3: this file should not be committed. Here is what needs to be done:

1. delete this file and commit the changes
2. add `scully.log` to `.gitignore` and commit again

projects/docs/src/assets/scully-routes.json
```diff
@@ -0,0 +1,33 @@
...
+    { "route": "/docs/textarea" },
+    { "route": "/docs/textboxgroup" },
+    { "route": "/docs/textlabel" },
+    { "route": "/docs/timeline" },
+    { "route": "/docs/timer" },
+    { "route": "/docs/toggle" },
+    { "route": "/docs/tooltip" },
+    { "route": "/docs/video" },
+    { "route": "/404" }
+]
```
yousifalraheem, line 33: I ran the `build:docs` on my machine, and this file seems to be generated every single time. Is this supposed to be in the `assets` folder? Shouldn't this file get dropped in the `dist` folder instead?

If it is supposed to be dropped here, can you at least run a prettier on the file after it's generated? because the generated file is minified and it will always show git changes when running a build.

scully.docs.config.ts
```diff
@@ -0,0 +1,14 @@
...
+import { baseHrefRewrite } from "@scullyio/scully-plugin-base-href-rewrite";
+
+const pkg = require("./package.json");
+const defaultPostRenderers = [baseHrefRewrite];
+setPluginConfig(baseHrefRewrite, { href: pkg.homepage });
+
+export const config: ScullyConfig = {
+    projectRoot: "./projects/docs/src",
+    projectName: "docs",
+    outDir: "./dist/static",
```
yousifalraheem, line 11: I noticed that the build is generating the docs into `dist/static`. This mean that the deployment of docs won't succeed.

You need to either change the directory name that Scully outputs to `docs`
```suggestion
    outDir: "./dist/docs",
```

Or change the publish directory in `.github/workflows/gh-pages.yml`. Currently it points to `dist/docs`
```yaml
    - name: GitHub Pages
      uses: crazy-max/ghaction-github-pages@v1
      with:
        build_dir: dist/docs
      env:
        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
```

projects/docs/src/app/docs-wrapper/docs-wrapper-routing.module.ts
```diff
@@ -0,0 +1,143 @@
+import { NgModule } from "@angular/core";
+import { Routes, RouterModule } from "@angular/router";
+import { DocsWrapperComponent } from "./docs-wrapper.component";
+import components from "../../assets/components-list.json";
+
+function getComponentPageRoutes(): Routes {
```
mario-subo, line 6: So this dynamic import doesn't work anymore?

projects/docs/src/assets/scully-routes.json
kherP: it will be generated in assets folder first then copied to dist folder. default scully function

projects/docs/src/app/docs-wrapper/docs-wrapper-routing.module.ts
kherP: removed

scully.docs.config.ts
kherP: rename to docs but we still another folder for `ng build` as scully run based on the built folder

scully.log
kherP: added

projects/docs/src/assets/scully-routes.json
yousifalraheem: then you can add a `post` npm hook to delete the folder
```bash
rm projects/docs/src/assets/scully-routes.json
```

or you can add it to `.gitignore`

PR #451, opened 2020-07-30
Refactor/modal

yousifalraheem:
Refactored Modal component to match the same logic and styles of the one in React Components library. Now it uses CSS instead of Angular animation library. Also fixed the hide animation and added an optional close button.


.vscode/settings.json
```diff
@@ -3,5 +3,8 @@
     "editor.formatOnSave": true,
     "[typescript]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode"
+    },
+    "[html]": {
+        "editor.defaultFormatter": "esbenp.prettier-vscode"
```
kherP, line 8: This change is intended to be committed?

projects/docs/src/app/docs-wrapper/doc-page/doc-page.component.html
```diff
@@ -1,25 +1,27 @@
...
         <ng-content select="[controls]" controls></ng-content>
     </app-doc-playground>
 
-    <app-doc-apis *ngIf="section === 1" [inputs]="inputs" [outputs]="outputs"></app-doc-apis>
+    <app-doc-apis *ngSwitchCase="1" [inputs]="inputs" [outputs]="outputs"></app-doc-apis>
 
-    <app-doc-notes *ngIf="contentNotes.innerText.length && section === 2">
-        <div [innerHTML]="contentNotes.innerHTML"></div>
-    </app-doc-notes>
+    <ng-container *ngIf="showNotes">
```
kherP, line 20: Why do you still need the if statement here? I think we can just modify the tab list

projects/ng-components/lib/modal/modal.module.ts
```diff
@@ -1,13 +1,10 @@
...
-import { SebModalBackdropModule } from "./modal.backdrop";
 import { CommonModule } from "@angular/common";
-import { ModalService } from "./modal.service";
 
 @NgModule({
-    imports: [CommonModule, SebModalBackdropModule],
+    imports: [CommonModule],
     declarations: [ModalComponent],
     exports: [ModalComponent],
-    providers: [ModalService],
```
kherP, line 11: Why delete the service?

.vscode/settings.json
yousifalraheem: yes

projects/docs/src/app/docs-wrapper/doc-page/doc-page.component.html
yousifalraheem: Because I don't want to be rendered at all if there are not notes

projects/ng-components/lib/modal/modal.module.ts
yousifalraheem: There is no more service, it's removed

projects/docs/src/app/docs-wrapper/doc-page/doc-page.component.html
yousifalraheem: fixed

PR #462, opened 2020-08-03
fix(radio-group): fixes focusing and custom template bugs

### LINKED ISSUE #443: selecting items in radio group within a scrollable view causes the view to "jump"
**Describe the bug**
If using the radio group in a scrollable view selecting different items will cause the view to move randomly possibly because of items being focused manually in the component after they are being selected

**To Reproduce**
1. Create a scrollable view
2. Add a radio group somewhere in the middle
3. Try clicking on different items of the radio group
4. see the view move

**Expected behavior**
the scroll of the parent view should not be affected

As a (temporary?) fix focusing on the items should be disabled

mario-subo:
fix #443 

While making the change I also noticed the custom template was not working at all since the key was removed so I fixed that as well. 

image


projects/ng-components/lib/radio-group/radio-group.component.ts
```diff
@@ -104,8 +104,10 @@ export class RadioGroupComponent implements ControlValueAccessor, AfterViewCheck
         if (!this.didFocus && this.list && this.list.length) {
             const currentFocused: number = this.list.findIndex(e => e && this.selectedValue && e.value === this.selectedValue.value);
             if (currentFocused > -1 && this.radioRefs.toArray()[currentFocused] && this.radioRefs.toArray()[currentFocused].nativeElement) {
-                this.radioRefs.toArray()[currentFocused].nativeElement.focus();
-                this.didFocus = true;
+                setTimeout(() => {
+                    this.radioRefs.toArray()[currentFocused].nativeElement.focus();
```
yousifalraheem, line 108: Can you just save guard this?
```suggestion
                    this.radioRefs.toArray()[currentFocused].nativeElement?.focus();
```

projects/ng-components/lib/radio-group/radio-group.component.ts
```diff
@@ -104,8 +104,10 @@ export class RadioGroupComponent implements ControlValueAccessor, AfterViewCheck
         if (!this.didFocus && this.list && this.list.length) {
             const currentFocused: number = this.list.findIndex(e => e && this.selectedValue && e.value === this.selectedValue.value);
             if (currentFocused > -1 && this.radioRefs.toArray()[currentFocused] && this.radioRefs.toArray()[currentFocused].nativeElement) {
-                this.radioRefs.toArray()[currentFocused].nativeElement.focus();
-                this.didFocus = true;
+                setTimeout(() => {
+                    this.radioRefs.toArray()[currentFocused].nativeElement?.focus();
+                    this.didFocus = true;
+                }, 0);
```
kherP, line 110: maybe we can use changesDetectionRef here instead of using setTimeout

mario-subo: I tried but changing the change detection strategy does not fix the scroll jumping issue ( re #443 ). SetTimeout is the only thing that fixes is unfortunately.

PR #496, opened 2020-08-17
refactor(textboxgroup): rename textboxgroup to textbox

### LINKED ISSUE #477: Textboxgroup should not render with `form-group` class
**Describe the bug**
Currently the textboxgroup will render with a `form-group` class which will add unnecessary `margin-bottom`.

yousifalraheem:
there will be no more separation between textbox and textboxgroup, instead, textbox will implement
the features of both

BREAKING CHANGE: Textboxgroup has been renamed. All instances where it's used need to be fixed to
address this change.

fixes #477


projects/ng-components/lib/accordion/accordion.component.spec.ts
```diff
@@ -7,7 +7,7 @@ import { DynamicStylePipe } from "./accordion-style.pipe";
 import { CommonModule } from "@angular/common";
 
 @Component({
-    selector: "test-sebng-textboxgroup",
+    selector: "test-sebng-textbox",
```
mario-subo, line 10: this is the accordion spec lol

PR #515, opened 2020-08-27
fix(dropdown): fixes bug with no results text not being displayed

mario-subo:


kherP (approved):

mario-subo:
@kherP should I merge?

PR #521, opened 2020-09-03
build(dep): update dependencies

kherP:


package.json
```diff
@@ -58,6 +58,7 @@
         "@angular/router": "~10.0.7",
         "@sebgroup/bootstrap": "^5.3.0",
         "@sebgroup/frontend-tools": "^2.0.1",
+        "bl": "^4.0.3",
```
yousifalraheem, line 61: This shouldn't be here! It's a buffer list for node. I'm not sure why you need it, but even if you need it, since it's a node dependency, it means it will be consumed during build time and it can be moved to `devDependencies`

yousifalraheem (changes requested):

yousifalraheem (approved):

PR #533, opened 2020-10-21
feat: toggle selector

omarVodiaK:


projects/ng-components/lib/toggle-selector/toggle-selector.component.html
```diff
@@ -0,0 +1,28 @@
...
+            (focus)="focus(idx)"
+            (blur)="blur()"
+            [defaultChecked]="item.selected"
+        />
+        <div class="icon-container" *ngIf="item.optionItem.icon" [innerHTML]="item.optionItem.icon | safeHtml"></div>
+        <label [attr.for]="item.optionItem.value" *ngIf="!item.optionItem.customLabel; customLabel">
+            {{ item.optionItem.label }}
+        </label>
+        <div #customLabel [innerHtml]="item.optionItem.customLabel"></div>
+    </div>
```
yousifalraheem, line 25: I think there is a chance of `item.optionItem` to not exist. Therefore, you either implement optional chaining in all the instances if `item.optionItem.x`. Or you add an `ngIf` to a container.

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            component.list = [
+                { value: "1", label: "Four mississipi" },
+                { value: "2", label: "Five mississipi" },
+            ];
+            fixture.detectChanges();
+        });
+
+        it("Should have type radio when multi is false", () => {
+            const elements: DebugElement[] = fixture.debugElement.queryAll(By.css(".toggle-selector input[type=radio]"));
+            expect(elements.length).toBe(component.list.length);
```
yousifalraheem, line 150: I think this is cleaner and easier to read
```suggestion
            expect(elements).toHaveLength(component.list.length);
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+        });
+
+        it("should not have anything checked", () => {
+            const inputEl: DebugElement[] = fixture.debugElement.queryAll(By.css("input"));
+            expect(inputEl.every(el => el.nativeElement.checked === true)).toBeFalsy();
```
yousifalraheem, line 155: The comparision here is redundunt.
```
(Checked = true) === true => true
(Checked = false) === true => false
```
You could just returned `checked`
```suggestion
            expect(inputEl.every(el => el.nativeElement.checked)).toBeFalsy();
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+        });
+
+        it("Should have only one checked when an option is clicked and move to another option if it's clicked", () => {
+            const element: HTMLElement[] = fixture.nativeElement.querySelectorAll(".toggle-selector");
+            element[1].dispatchEvent(new Event("click"));
+            fixture.detectChanges();
+            const inputEl: DebugElement[] = fixture.debugElement.queryAll(By.css("input"));
+            expect(inputEl.every(el => el.nativeElement.checked === true)).toBeFalsy();
```
yousifalraheem, line 163: Same here, just return `checked`
```suggestion
            expect(inputEl.every(el => el.nativeElement.checked)).toBeFalsy();
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            expect(inputEl[1].nativeElement.checked).toBeTruthy();
+
+            element[0].dispatchEvent(new Event("click"));
+            fixture.detectChanges();
+            expect(inputEl.every(el => el.nativeElement.checked === true)).toBeFalsy();
```
yousifalraheem, line 168: Same here, just return `checked`
```suggestion
            expect(inputEl.every(el => el.nativeElement.checked)).toBeFalsy();
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            fixture.detectChanges();
+        });
+
+        it("Should have type radio when multi is false", () => {
+            const elements: DebugElement[] = fixture.debugElement.queryAll(By.css(".toggle-selector input[type=radio]"));
+            expect(elements.length).toBe(component.list.length);
+        });
+
+        it("should not have anything checked", () => {
+            const inputEl: DebugElement[] = fixture.debugElement.queryAll(By.css("input"));
```
yousifalraheem, line 154: an easier way to do this check is by querying like this:
```suggestion
            const inputEl: DebugElement[] = fixture.debugElement.queryAll(By.css("input:checked"));
            expect(inputEl).toHaveLength(0);
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            const element: HTMLElement[] = fixture.nativeElement.querySelectorAll(".toggle-selector");
+            element[1].dispatchEvent(new Event("click"));
+            fixture.detectChanges();
+            const inputEl: DebugElement[] = fixture.debugElement.queryAll(By.css("input"));
+
+            expect(inputEl.every(el => el.nativeElement.checked === true)).toBeFalsy();
+
+            element[1].dispatchEvent(new Event("click"));
+            fixture.detectChanges();
+            expect(inputEl.some(el => el.nativeElement.checked === true)).toBeFalsy();
```
yousifalraheem, line 134:
```suggestion
            expect(inputEl.every(el => el.nativeElement.checked)).toBeTruthy();
        });

        it("should uncheck when previously checked option is clicked", () => {
            const element: HTMLElement[] = fixture.nativeElement.querySelectorAll(".toggle-selector");
            element[1].dispatchEvent(new Event("click"));
            fixture.detectChanges();
            const inputEl: DebugElement[] = fixture.debugElement.queryAll(By.css("input"));

            expect(inputEl.every(el => el.nativeElement.checked)).toBeFalsy();

            element[1].dispatchEvent(new Event("click"));
            fixture.detectChanges();
            expect(inputEl.some(el => el.nativeElement.checked)).toBeFalsy();
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    it("Should disable the element when disabled prop is set to true", () => {
+        component.disabled = true;
+        fixture.detectChanges();
+        const container: DebugElement = fixture.debugElement.query(By.css(".disabled"));
+        const elements: DebugElement[] = fixture.debugElement.queryAll(By.css(".toggle-selector input:disabled"));
+        expect(container).toBeDefined();
+        expect(container).not.toBeNull();
+        expect(elements).toBeDefined();
+        expect(elements).toBeTruthy();
+        expect(elements.length).toBe(component.list.length);
```
yousifalraheem, line 77:
```suggestion
        expect(container).not.toBeNull();
        expect(elements).toHaveLength(component.list.length);
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    error: boolean;
+    errorMessage: string;
+
+    constructor() {
+        this.list = [
+            { value: "1", label: "Four mississipi", disabled: true },
+            { value: "2", label: "Five mississipi" },
+        ];
+    }
+}
```
yousifalraheem, line 35: I think if you extend `ToggleSelectorComponent`, then you won't need to redfined the props. Can you try that?
```suggestion
class ToggleSelectorTestComponent extends ToggleSelectorComponent {
    @ViewChild("toggleSelector") toggleSelector: ToggleSelectorComponent;

    constructor() {
        this.multi = this.disabled = false;
        this.list = [
            { value: "1", label: "Four mississipi", disabled: true },
            { value: "2", label: "Five mississipi" },
        ];
    }
}
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.html
```diff
@@ -0,0 +1,28 @@
...
+        (click)="handleItemOnClick($event, idx)"
+    >
+        <input
+            [type]="inputType"
+            [attr.aria-label]="item.optionItem.label"
+            [id]="item.optionItem.value"
+            [name]="name"
+            [disabled]="disabled || item.optionItem.disabled"
+            (focus)="focus(idx)"
+            (blur)="blur()"
```
yousifalraheem, line 17: Wouldn't this be easier to maintain? since it's just a simple operation we don't need to go back and forth between the template and the javascript part.
```suggestion
            (focus)="focusedItem = idx"
            (blur)="focusedItem = null"
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.html
```diff
@@ -0,0 +1,28 @@
...
+    <div
+        class="ac toggle-selector"
+        [class.error]="error"
+        [class.checked]="item.selected"
+        [class.disabled]="disabled || item.optionItem.disabled"
+        [class.focus]="focusedItem === idx"
+        (click)="handleItemOnClick($event, idx)"
+    >
+        <input
+            [type]="inputType"
```
yousifalraheem, line 11: I don't think there is a need for a getter here. It's quite straightforward
```suggestion
            [type]="multi ? 'checkbox' : 'radio'"
```

yousifalraheem (changes requested):

projects/ng-components/lib/toggle-selector/toggle-selector.component.html
omarVodiaK: there is no possibility for it to happen and here is why:
the template you see now is inside a loop that uses an array called disaplyList of type:
```typescript
interface IDisplayToggleSelector extends IToggleSelector {
    optionItem: IToggleSelector;
    selected: boolean;
}

export interface IToggleSelector {
    value: string;
    label?: string;
    icon?: string;
    iconPosition?: IconPosition;
    customLabel?: HTMLElement;
    description?: string;
    disabled?: boolean;
}
```

displayList Array is populated when the @input list of the component is set, there for it always have optionItem object.

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
omarVodiaK: This throws an error Property 'toHaveLength' does not exist on type 'ArrayLikeMatchers<DebugElement>'

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
omarVodiaK: This throws an error Property 'toHaveLength' does not exist on type 'ArrayLikeMatchers'

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
omarVodiaK: Im not redefining the Inputs of the components, those properties are variables that needs to be passed to the test comoponent and they are useful for passing different values through-out the test.

```typescript
@Component({
    selector: "test-sebng-toggle-selector",
    template: `<sebng-toggle-selector
        #toggleSelector
        [multi]="multi"
        [list]="list"
        [(ngModel)]="model"
        [disabled]="disabled"
        [error]="error"
        [errorMessage]="errorMessage"
    ></sebng-toggle-selector>`,
})
```

projects/ng-components/lib/toggle-selector/toggle-selector.component.html
omarVodiaK: let's not write JS operation inside HTML files, it get's very confusing when the component is huge.

projects/ng-components/lib/toggle-selector/toggle-selector.component.html
omarVodiaK: (Russian accent) but but but i created this type: 
```typescript
export type InputType = "checkbox" | "radio";
```

mario-subo (approved):

projects/ng-components/lib/toggle-selector/toggle-selector.component.spec.ts
yousifalraheem: Same usage, if you extend it, it will already be there. I use that technique a lot. Try it la

projects/ng-components/lib/toggle-selector/toggle-selector.component.html
yousifalraheem: this is very similar to:
```html
<div [ngClass]="{ isDisabled: disabled, theme: theme || 'primary' }"></div>
```
The `ng-class` directive processes JavaScript logic

yousifalraheem (approved):

PR #600, opened 2021-12-13
ALFA-86: Added automation id

NiklasBack:


kherP (commented):
Hi Niklas, thank you for making this PR. Is it possible to refactor the code to append the attributes more dynamically?

For instance, introduce a new prop on dropdown level, and append the value of the new prop to the label element with some helps from either a custom directive or viewinit hook. https://stackoverflow.com/questions/44583258/angular-2-dynamic-bind-attribute-name/44606040

